PR #108, opened 2015-01-31
Generic Query with Join

tuliren:
# Generic Query

## Introduction
This is a query system based on db connection instead of individual model persistence. Consequently, it supports join operation among any model within the database.

## Example
```java
// create query
Records results = GenericQuery.create(DATABASE_CONNECTION1)
    .from(User.TBL)
    .leftJoin(Comment.TBL).on(User.ID, Comment.COMMENTER_ID)
    .where(Comment.CONTENT.isNotNull())
    .orderBy(User.NUM_POSTS, DESC)
    .orderBy(Comment.ID)
    .select(User.ID, User.BIO, User.HANDLE, User.NUM_POSTS, Comment.ID, Comment.COMMENTER_ID, Comment.CONTENT)
    .fetch();

// retrieve the queried fields
for (Record record : results) {
  String bio = record.getString(User.BIO);
  Integer id = record.getIntFromLong(User.ID);
  Integer numPosts = entry.getInt(User.NUM_POSTS);
}
```

### Column
`Column` represents a column in a table. It has a reference to the name and alias of the table, a field (`_Field`) within that model of the table, and the field type. An object of `Column` is the basic element the user needs to specify for most query clauses. The type is used for runtime type-safety check.

### Record
A `Record` represents a collection of columns specified by the select clause. The returned result from the generic query is a `List` of `Record`. The user can get the corresponding value of a column by a group of `get<type>` methods (e.g. `getInt`, `getLong`, `getByteArray`). The type must match with that of the specified `Record`. Otherwise, a runtime exception will be thrown.

There is an extra `getIntFromLong` method that encapsulates a safe `Integer` to `Long` conversion. No need to call `Rap.safeLongToInt`.

### Records
The `Records` class encapsulates the returned query result list. The user can iterate the records by calling its iterator. The `get<type>s` methods extracts a specific column from all the records into a list.

### Query Constraints
If no constraint is provided, the query will by default return all records of the selected table.

This query system reuses all the query constraints as the per-model query system. Two changes are added to these constraints:
- All constraints now implements the `QueryCondition` interface, which has only one method `getSqlStatement`. This is to ensure that the generic query system can easily construct a sql statement.
- The `Enum field` variable is replaced with a `Column`. This is because for generic queries, both the table and field name need to be specified.

## Additional Changes
- `ModelWithId` now has a `getIntId` method.


tuliren:
@seancarr This PR is ready for review. Can you take a look? Thanks.

seancarr:
@tuliren, you should add the documentation from this pull request to the README. It's very helpful.

src/java/com/rapleaf/jack/ModelField.java
```diff
@@ -0,0 +1,61 @@
...
+  public Enum getField() {
+    return field;
+  }
+
+  public Class getType() {
+    return type;
+  }
+
+  public String getSqlKeyword() {
+    String fieldKeyword = field != null ? field.toString() : DEFAULT_ID_FIELD;
```
seancarr: Is "null" for id just necessary because of how we use ModelWithId and it isn't an Attribute? Do you feel this is something we should change at some point? Ideally I'd like to have ModelWithId just be a specific type of Model and many usages, like this one, wouldn't need to rely on the WithId part.

tuliren:
Will do that when this PR is finalized. The description here is kind of verbose and I image that a few things may be changed.

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,149 @@
...
+
+public class GenericQuery {
+
+  private final BaseDatabaseConnection dbConnection;
+  private Class<? extends ModelWithId> mainModel;
+  private List<JoinCondition> joinConditions;
+  private List<WhereConstraint> whereConstraints;
+  private List<OrderCriterion> orderCriteria;
+  private Set<ModelField> selectedIModelFields;
+  private Optional<LimitCriterion> limitCondition;
```
seancarr: I know you want to change these names later but for now should this be called limitCriteria to stick with the pattern of following the current class names?

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,149 @@
...
+    StringBuilder statement = new StringBuilder();
+
+    statement.append(getSelectClause())
+        .append(getJoinClause())
+        .append(getWhereClause());
+
+    if (isOrderedQuery()) {
+      statement.append(getOrderClause());
+      if (!orderCriteria.isEmpty()) {
+        statement.append(getLimitClause());
```
seancarr: Why is limit only used when it's an ordered query? I may want to limit a query without specifying an order.
Also, aren't isOrderedQuery() and "if (!orderCriteria.isEmpty()) {" redundant of each other?

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,149 @@
...
+import com.rapleaf.jack.BaseDatabaseConnection;
+import com.rapleaf.jack.ModelField;
+import com.rapleaf.jack.ModelWithId;
+
+public class GenericQuery {
+
+  private final BaseDatabaseConnection dbConnection;
+  private Class<? extends ModelWithId> mainModel;
+  private List<JoinCondition> joinConditions;
+  private List<WhereConstraint> whereConstraints;
```
seancarr: Some of these could be final as well, right? You never allow someone to replace the original list with a new list.

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,149 @@
...
+  private List<OrderCriterion> orderCriteria;
+  private Set<ModelField> selectedIModelFields;
```
seancarr: Why the "I"? The objects are ModelFields not IModelFields. Was this maybe a refactor at some point?

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,149 @@
...
+  private Optional<LimitCriterion> limitCondition;
+
+  private GenericQuery(BaseDatabaseConnection dbConnection) {
```
seancarr: Do you plan on adding GROUP BY functionality later? Related to this would be the ability to select functions like MAX, COUNT, etc. Those would probably be some type of generic ModelField that is just a String, not an Enum.
I'm not suggesting that functionality should be part of this version but it's probably important to consider now since it might impact the design of things like ModelField.

src/java/com/rapleaf/jack/ModelField.java
tuliren: Yes. It's `null` because `id` is not an attribute. I do feel that `id` has been given too many speciality. A few query methods can be simpler if it is just an ordinary field like the others.

src/java/com/rapleaf/jack/queries/GenericQuery.java
tuliren: Is that without `ORDER BY` clause, the returned result is a set of records with no particular order, so the same query with a `LIMIT` clause may return different results?

Sorry about the second part. They are redundant.

src/java/com/rapleaf/jack/queries/GenericQuery.java
tuliren: Sorry that I totally neglected the `GROUP BY` previously. I think there is a way to implement it without much refactoring. Will work on it this week.

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,147 @@
...
+  String getSqlStatement() {
+    StringBuilder statement = new StringBuilder();
+
+    statement.append(getSelectClause())
+        .append(getJoinClause())
+        .append(getWhereClause());
+
+    if (isOrderedQuery()) {
+      statement.append(getOrderClause());
+      statement.append(getLimitClause());
```
seancarr: Do you want to chat about this offline? I feel somewhat strongly that we should support limit without order. Also, if we didn't we should raise an exception if there was a limit so people know their input is being ignored.

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+      }
+    }
+
+    String tableName = Utility.getTableName(mainModel);
+    return clause.append(" FROM ").append(tableName).append(" ").toString();
+  }
+
+  private String getJoinClause() {
+    return getClause(joinConditions, "", " ");
```
seancarr: This implementation doesn't support self-joins where the same table appears multiple times. In that case you need to be able to rename tables you are selecting from. Similarly we might want to be able to rename columns we select (at least once we add support for group by).

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private String getWhereClause() {
+    return getClause(whereConstraints, "WHERE ", " AND ");
```
seancarr: How would ORs be supported?

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,147 @@
...
+      return "";
+    }
+  }
+
+  private <T extends IQueryCondition> String getClause(Collection<T> conditions, String initialKeyWord, String separator) {
+    if (conditions.isEmpty()) {
+      return "";
+    }
+
+    StringBuilder clause = new StringBuilder(initialKeyWord);
```
seancarr: I think KeyWord should be Keyword

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,165 @@
...
+    return this;
+  }
+
+  public GenericQueryBuilder limit(int offset, int limit) {
+    genericQuery.addLimitCondition(new LimitCriterion(offset, limit));
+    return this;
+  }
+
+  public GenericQueryBuilder limit(int limit) {
+    genericQuery.addLimitCondition(new LimitCriterion(0, limit));
```
seancarr: why use the 2 arg constructor here? Just passing limit seems cleaner.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,165 @@
...
+        if (++retryCount > MAX_CONNECTION_RETRIES) {
+          throw new IOException(e);
+        }
+      } catch (SQLException e) {
+        throw new IOException(e);
+      }
+    }
+  }
+
+  private void queryExecution(List<QueryEntry> results, PreparedStatement preparedStatement) throws SQLException {
```
seancarr: Since these are all select queries and always return results why have it be void and populate the list passed in instead of returning the results List? I actually think this could cause issues with the retry logic if the exception happened after results was partially populated.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,165 @@
...
+      } catch (SQLRecoverableException e) {
+        dbConnection.resetConnection();
+      } catch (SQLException e) {
+      }
+    }
+  }
+
+  private QueryEntry parseResultSet(ResultSet queryResultSet) throws SQLException{
+    Set<ModelField> selectedModelFields = genericQuery.getSelectedModelFields();
+    if (selectedModelFields.isEmpty()) {
```
seancarr: Won't this be the case when we want to select all columns from a table?

src/java/com/rapleaf/jack/queries/Utility.java
```diff
@@ -0,0 +1,14 @@
+package com.rapleaf.jack.queries;
+
+import com.rapleaf.jack.ModelWithId;
+
+public class Utility {
+
+  public static String getTableName(Class<? extends ModelWithId> model) {
+    try {
+      return (String)model.getDeclaredField("_TABLE_NAME").get(null);
```
seancarr: Do we want to expose this in a cleaner way?

test/java/com/rapleaf/jack/TestGenericQuery.java
```diff
@@ -0,0 +1,159 @@
...
+import com.rapleaf.jack.test_project.database_1.iface.ICommentPersistence;
+import com.rapleaf.jack.test_project.database_1.iface.IPostPersistence;
+import com.rapleaf.jack.test_project.database_1.iface.IUserPersistence;
+import com.rapleaf.jack.test_project.database_1.models.Comment;
+import com.rapleaf.jack.test_project.database_1.models.Post;
+import com.rapleaf.jack.test_project.database_1.models.User;
+
+import static org.junit.Assert.*;
+
+public class TestGenericQuery {
```
seancarr: As part of making some of the changes discussed above you should add tests for those specific cases so it's clear what our intention was / we make sure they work.

seancarr:
Sorry for the slow review. Done now.

tuliren:
Thanks for taking the time. The comments this time are all tough ones.

For aggregated functions, I think we don't need to rename the column. My design is to create a class `AggregatedField` that derives from `ModelField`. An aggregrated field keeps track of the aggregation function, and is responsible for generating the correct name for the query. It can be created by a static call on any of the `ModelField` instance, e.g. `MAX(User.ID)`, in which `MAX` is the static method, and `User.ID` is a `ModelField`. In this way, the user can create an aggregation just like in SQL syntax.

src/java/com/rapleaf/jack/queries/GenericQuery.java
tuliren: I will add support for `limit` without order. But the result may be non-deterministic. That's my concern. Currently working on self-join, group by and custom table alias (that's a must for self-join). The current design of self-join syntax is like this:

```java
.from(Employee.class)
.leftJoin(Employee.class).as("manager").on(Employee.NAME.of("manager"), User.MANAGER_NAME)
```
So in this table, an employee can be the manager of another employee. The `Employee` table is aliased to `manager` in the join, and the join condition is `employee.manager_name = manager.name`.

When selecting a field from a self-joined table, the user needs to specify which table he / she wants (if that table is aliased).

```java
.select(Employee.NAME.of("manager"), Employee.NAME)
```
The above selection would give the name of the manager and the one he / she manages. The first `ModelField` is from the aliased `employee` table. The second one is from the non-aliased `employee` table.

The chained `.leftJoin().as().on()` structure is not necessary though. All these parameters can be passed into one `leftJoin` call as it is now. Probably the former is more readable.

seancarr: All sounds good. I don't have a strong preference between leftJoin().as().on() vs 3 args to 1 function. Definitely does read nicely.

For limit without order I think we should allow it because mysql does and rails does. Nowhere does it say that an unordered query will be deterministic and this is no different. In practice though it is deterministic, but yes we shouldn't rely on that.

tuliren:
@seancarr 
All done. It supports most SQL clauses except for `HAVING`.

The major changes are:
1) Each model has a static inner class `Table`, which encapsulates all the fields and support table alias. The user can create an aliased table by `User.Table developers = User.Table.as("developers")`, and then get the column "developers.handler" by `developers.HANDLER`.
2) `ModelField` is replaced by `Column` because it not longer keeps a reference of the model, but a reference of a `Table`, so it is more like a column now.
3) More test cases are added. I replicated a lot of the tests from model queries to make sure they also work for the generic query.

I am also planning to add the `HAVING` clause. But the basic structure does not need to be changed.

src/java/com/rapleaf/jack/AbstractTable.java
```diff
@@ -0,0 +1,30 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Sets;
+
+public class AbstractTable implements Table {
+  protected final String table;
+  protected final String alias;
+  protected final Set<Column> allColumns;
+
+  protected AbstractTable(String table, String alias) {
+    Preconditions.checkArgument(table != null && !table.isEmpty(), "Table name cannot be null or empty.");
```
seancarr: Since you're already using Preconditions how about use Strings.isNullOrEmpty()?

src/java/com/rapleaf/jack/Column.java
```diff
@@ -0,0 +1,72 @@
...
+    this.type = type;
+  }
+
+  protected Column(Column that) {
+    this.table = that.table;
+    this.field = that.field;
+    this.type = that.type;
+  }
+
+  public static Column fromKey(String table) {
```
seancarr: Do we refer to the primary key id as "key" elsewhere? If so, that's fine but if not we should think about this naming.

src/java/com/rapleaf/jack/Column.java
```diff
@@ -0,0 +1,72 @@
...
+    return new Column(table, null, Long.class);
```
seancarr: I like how this sets us up for a better path to having some primary keys be Integer vs Long :)

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+import com.rapleaf.jack.Table;
+import com.rapleaf.jack.queries.where_operators.IWhereOperator;
+
+public class GenericQueryBuilder {
+  private static int MAX_CONNECTION_RETRIES = 1;
+
+  private final GenericQuery genericQuery;
+  private final BaseDatabaseConnection dbConnection;
+
+  public GenericQueryBuilder(BaseDatabaseConnection dbConnection, GenericQuery genericQuery) {
```
seancarr: Something about the from() being public on the GenericQuery and everything else being on the builder feels weird to me. The idea behind it is that you can only have one from() but many of everything else? I think it would be nice for a user to only need the builder to build a query. I don't have clear ideas about how to change things and maybe they're fine this way but it does feel a little strange.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+  public JoinConditionBuilder innerJoin(Table table) {
+    return new JoinConditionBuilder(this, JoinType.INNER_JOIN, table);
+  }
+
+  GenericQueryBuilder addJoinCondition(JoinCondition joinCondition) {
+    genericQuery.addJoinCondition(joinCondition);
+    return this;
+  }
+
+  public GenericQueryBuilder where(Column column, IWhereOperator operator) {
```
seancarr: Is there a way to group where conditions so you can do "A AND (B OR C)"? What happens if someone uses the null logic method multiple times?

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,176 @@
...
+    } finally {
+      try {
+        if (queryResultSet != null) {
+          queryResultSet.close();
+        }
+        preparedStatement.close();
+      } catch (SQLRecoverableException e) {
+        dbConnection.resetConnection();
+      } catch (SQLException e) {
+        // ignore
```
seancarr: It would be a good idea to incorporate some of the logging improvements that Tenzing added in his PRs today.

seancarr:
@tuliren those are all my comments for this round.

src/java/com/rapleaf/jack/Column.java
tuliren: Oh, right. We don't do that anywhere else. So maybe `fromId` or `fromPrimaryKeyId`?

seancarr: I think `fromId` works since we generally refer to it as just "id" elsewhere too

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  public String getFullSqlStatement() throws IOException {
+    return getPreparedStatement().toString();
+  }
+
+  public Records fetch() throws IOException {
+    int retryCount = 0;
+    PreparedStatement preparedStatement = getPreparedStatement();
+    LOG.info("query: {}", preparedStatement.toString());
```
tuliren: Do you think this will make the log too verbose? Maybe important for debugging though.

seancarr: Yeah, it'll be way too verbose. Just make it debug level and it won't be logged by default.

src/java/com/rapleaf/jack/Column.java
tuliren: If we are going to have `Integer` primary keys in the future, the method may be renamed as `fromLongId`, so that later on we can add a `fromIntId` counterpart.

seancarr: I think we can leave it as is for now since there will be so many other changes to support Integer primary keys.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
tuliren: This is one of the places where I struggled the most. You are right that it will be a cleaner design to have the `GenericQueryBuilder` simply build the query. And the `GenericQuery` executes the query.

So maybe something like this:
`GenericQueryBuilder qb = GenericQueryBuilder.from().where(...);`
`Records records = GenericQuery.execute(qb);`

The public static `from()` will be the only method that returns a `GenericQueryBuilder` which ensures only one `FROM` clause for each query.

seancarr: I like that idea. Why not go even further and have an `execute()` method on `GenericQueryBuilder` too? It's similar to the query builder stuff you added for the models where they build the query and then call `find()`.

tuliren: Got it.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
tuliren: Any `WHERE` clause with a null logic other than the first one will default to `AND`. I know this is not great.

Putting everything into only one `WHERE` clause can solve both problems. I can think of two approaches:
1. The OOP way: `where(a.and(b.or(c)))`;
2. The LISP way: `where(AND(a, OR(b, c)))`.

In both approaches, `a = User.ID.equalsTo(123L)`, `b = User.HANDLER.equalsTo("abc")` and `c = User.BIO.equalsTo("abc")`, in which `equalsTo` is a method called on a `Column` and returns a where constraint.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
tuliren: I remembered why I did not adopt this approach in the first place. The builder need to take two parameters: a connection, and a primary table. And I don't want to put them in the same method. So just using something like `GenericQueryBuilder.from().where(...).execute` does not solve this issue. I will try to create one derived query builder class for each database which contains the connection by default. For our use case, the builder may be `RlDbQueryBuilder`.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
seancarr: The rails query builder only supports ANDs between multiple `where()` calls and then allows whatever sort of custom and/or combinations you want within a single `where()`. I think that let's us keep the interface simple (just one where method) but not necessarily limit it to just one call.

Between the two options you provide I prefer the OOP way because it reads the same way the sql or rails query would instead of being prefix notation. How nasty would that be to change?

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
seancarr: That makes sense.

src/java/com/rapleaf/jack/queries/GenericQueryBuilder.java
tuliren: I don't know whether it will be nasty or not yet, but that's fine. There must be a way.

The OOP approach has a big drawback. If there are lots of conditions, there will be a large number of parentheses:
```java
.where((a.or(b)).and(    // [a OR b] AND [c OR d] AND [e OR (f AND g)]
       (c.or(d)).and(
       (e.or(f, g)))))
```

I can combine the rails way into a single `WHERE` clause. Constraints separated by comma within the `where()` method have logic `AND`. Within each constraint, any combinations of `OR` and `AND` relations are allowed. This can significantly reduce the number of parentheses, and it seems easier to implement.

```java
.where(a.or(b),          // [a OR b] AND [c OR d] AND [e OR (f AND g)]
       c.or(d),          // AND logic is always represented by comma
       e.or(f, g))
```

seancarr: I don't think that's too many parens. That's roughly how many parens you'd need if you were writing the SQL too.
Having multiple arguments ANDed together sounds good. The most common queries would be ones like `where(a, b, c)` which is nice and clean.

tuliren:
@seancarr All done.

src/java/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -0,0 +1,288 @@
...
+    while (it.hasNext()) {
+      clause.append(it.next().getSqlStatement());
+      if (it.hasNext()) {
+        clause.append(separator);
+      }
+    }
+
+    return clause.append(" ").toString();
+  }
+}
```
tuliren: `getClauseFromColumns` and `getClauseFromQueryConditions` are basically the same except for the type of elements in the collection. I thought about making `Column` and other query condition classes implement the same interface with a `getSqlStatement` method to unify them, but that involves other namings issues. I'd like to leave that to a sweep ramp ticket.

seancarr:
Cool, looks good :)

PR #109, opened 2015-03-05
Remove mocks

pwestling:
Removes all of the mock implementations. Reasons:

1. They're already broken in that not all results were consistent with the actual DB
2. As we expand the query builder and it's usages, the mock db's will be useful in fewer and fewer situations 
3. You can get almost the same speed as the in-memory mock db by having test dbs in an in-memory sql isntance
4. Mock classes add a ton of PermGen usage to a project for almost no benefit

@seancarr @bpodgursky 

@eddiesiegel and @bryanduxbury: If you guys are still using Jack, let us know if you're using this functionality heavily. We'd encourage you to not use it since it doesn't really work correctly in all cases, and we're definitely not going to make the new queries work with the mocks, but if this is a super breaking change for you guys we might be able to figure something out.


bpodgursky:
beautiful

eddiesiegel:
We do use Jack, but we don't use the mock implementations, so the cleanup is appreciated :-)

Thanks for the heads up! :+1:

PR #110, opened 2015-03-05
make datetime fields not depend on user's time settings

pwestling:
prevents causing conflicts when you change timezones

@bpodgursky


pwestling:
@seancarr

PR #111, opened 2015-03-12
Remove Mock (continued)

tuliren:
Some of the mock code was not completely removed.

The deleted code in this PR includes:
1) The `AbstractMockDatabaseModel` class;
2) The `IWhereOperator#apply` method, which was only used in mock queries.


tuliren:
@seancarr By the way, this is a simple PR. It's ready to merge.

PR #112, opened 2015-03-19
adding slf4j/log4j maven dependencies

twshaw3:
@seancarr 

I don't think jack has the capability to log right now, so we'll need to add this to begin with?


seancarr:
Yep, looks good.

PR #113, opened 2015-03-19
passing the cause of the recoverable exception to resetConnection

twshaw3:
@seancarr 

I thought it might be cleanest for resetConnection() to optionally take a cause. Open to suggestions, though.


seancarr:
Makes sense and looks good.

PR #114, opened 2015-03-19
logging recoverable sql exceptions

twshaw3:
@seancarr 

I think this should log most of what people need to investigate root causes in case of connection reset attempts.

I'm also thinking of adding more logging to AbstractDatabaseModel, since there are cases when we ignore SqlExceptions in finally clauses.


seancarr:
Looks good.

PR #115, opened 2015-03-19
adding a helper method to deal with common cleanup and adding more loggi...

twshaw3:
...ng

@seancarr 

Not sure if there's a reason not to have a helper method here? Also, I'm not particularly happy with the name.


seancarr:
No reason.
The name and arguments are in a different order.
Maybe a name related to the context of these two things? closeQuery()? Don't have any great ideas.

twshaw3:
I like closeQuery; going with that.

PR #117, opened 2015-03-30
Fix Date and Datetime Query

tuliren:
@bpodgursky found a bug in queries with `Datetime` columns. The bug also applies to `Date` columns.

The issue is the same as [this StackOverflow post](http://stackoverflow.com/questions/10888194/mysql-select-timestamps-between-a-and-b-returning-all-or-0-timestamps). Querying date / datetime columns with long values won't always give the correct result because SQL engine expects date literal for those columns.

Since Jack treats SQL `Datetime` and `Date` as `Long` in Java, the query cannot auto apply the appropriate operations on the long variable. The fix here uses static functions to convert the long value(s) into date or datetime literal(s). For a long `Datetime`, a `java.util.Date` type is constructed and then formatted. For a long `Date`, a `java.sql.Date` type is constructed and converted as a String.

Query for `Date`:
```java
import static com.rapleaf.jack.queries.Functions.DATE;

results1 = Database1Query
    .from(User.TBL)
    .where(User.SOME_DATE.equalTo(DATE(longDateValue)))
    .select(User.ID, User.SOME_DATE)
    .fetch();
```

Query for `Datetime`:
```java
import static com.rapleaf.jack.queries.Functions.DATETIME;

results1 = Database1Query
    .from(User.TBL)
    .where(User.SOME_DATETIME.equalTo(DATETIME(longDatetimeValue)))
    .select(User.ID, User.SOME_DATETIME)
    .fetch();
```

I have considered several options before implementing this fix. The idealist solution may be to treat `Date` and `Datetime` as is instead of representing them as `Long` so that the query builder can perform the conversion automatically, but that will have significant impacts on the current use cases. Using static methods are not the cleanest way, but I imagine that there will be more supports for SQL functions in the future in Jack, and this design is consistent with it.

@seancarr

### DIFF
--- a/src/java/com/rapleaf/jack/queries/Functions.java
+++ b/src/java/com/rapleaf/jack/queries/Functions.java
@@ -1,0 +1,54 @@
+package com.rapleaf.jack.queries;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Collection;
+import java.util.List;
+
+import com.google.common.collect.Lists;
+
+public class Functions {
+
+  private static final DateFormat datetimeFormat = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss");
+
+  public static String DATETIME(long datetime) {
+    return datetimeFormat.format(new java.util.Date(datetime));
+  }
+
+  public static Collection<String> DATETIMES(long firstDatetime, long... otherDatetimes) {
+    List<String> timestamps = Lists.newArrayList(DATETIME(firstDatetime));
+    for (long datetime : otherDatetimes) {
+      timestamps.add(DATETIME(datetime));
+    }
+    return timestamps;
+  }
+
+  public static Collection<String> DATETIMES(Collection<Long> datetimes) {
+    List<String> timestamps = Lists.newArrayList();
+    for (Long datetime : datetimes) {
+      timestamps.add(DATETIME(datetime));
+    }
+    return timestamps;
+  }
+
+  public static String DATE(long date) {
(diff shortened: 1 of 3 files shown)


bpodgursky:
How hard would it be to throw a runtime error if you accidentally try to use a raw long as a query instead of using the static methods?  It would be good to actively let people know that their query won't work as expected.

src/java/com/rapleaf/jack/queries/Functions.java
```diff
@@ -0,0 +1,54 @@
...
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Collection;
+import java.util.List;
+
+import com.google.common.collect.Lists;
+
+public class Functions {
+
+  private static final DateFormat datetimeFormat = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss");
```
bpodgursky: SimpleDateFormat is not threadsafe -- should use DateTimeFormatter from Joda time.

tuliren:
I think it is not easy now. `Datetime` columns are typed as `Long` in Jack currently, and there is no proper way to differentiate it from an ordinary `Long` column.

tuliren:
How about `FastDateFormat`? It's thread-safe and in `org.apache.commons.lang.time`, which is already one of the dependencies.

bpodgursky:
I'm ok adding Joda time as a dependency, I think it's very lightweight.  But whichever you think is better.

tuliren:
It seems that Jodatime has a better performance over Apache common datetime. Switched.

PR #118, opened 2015-04-06
Generic Query SP1

tuliren:
@bpodgursky This PR adds support for the three features you talked about during the Hackweek.

### Connection Setup
The user can now set and reset the connection, or get the default connection from the implementation of the database.
```java
// create a generic query with custom connection
GenericQuery query = GenericQuery.create(conn).from(...)...

// create a generic query with default connection
IRlDb rldb = new DatabasesImpl().getRlDb();
GenericQuery query = rldb.createQuery().from(...)...
```

`GenericQuery#create` takes the custom connection and returns a `GenericQuery.Builder`. `GenericQuery.Builder#from` takes the first table for query and returns a `GenericQuery`. The reason that I have to use a `Builder` is because a query takes a connection and a basic table, and it's cleaner to put them in two methods. The previous `RlDbQuery` class has been removed, because it does not support custom connection reset. I will fix the use cases in the code base once it's merged.

### Generalized Join Condition
The join syntax is more generalized:
```java
rldb.createQuery()
    .from(User.TBL)
    .join(Post.TBL)
    .on(User.ID.equalTo(Post.USER_ID),
        Post.USER_ID.greaterThan(3).or(Post.USER_ID.equalTo(1)))
    .fetch();
```
So the join condition is exactly like a where clause. `AND` conditions are separated by comma, and the `OR` logic is specified by the `or` method.

### Column Constraints
Any constraint can take not only a constant, but also a `Column` now. The above example already showed this feature.

By combining these new features, it is now possible to carry out the query for an entry with a largest field within a group as suggested [here](http://stackoverflow.com/questions/7745609/sql-select-only-rows-with-max-value-on-a-column):
```java
User.Tbl temp = User.Tbl.as("temp");
Records records = db.createQuery()
    .from(User.TBL)
    .leftJoin(temp)
    .on(User.SOME_BOOLEAN.equalTo(temp.SOME_BOOLEAN),
        User.NUM_POSTS.lessThan(temp.NUM_POSTS))
    .where(temp.NUM_POSTS.isNull())
    .orderBy(User.HANDLE)
    .select(User.HANDLE, User.SOME_BOOLEAN, User.BIO, User.NUM_POSTS)
    .fetch();
```
This query gets the user entry with the largest `NUM_POSTS` within each of the two `SOME_BOOLEAN` group.

The only thing still missing is the method that returns a Jack model from a `Record`. Since the new cycle starts tomorrow, I'm afraid I won't work on it sometime soon.


src/java/com/rapleaf/jack/queries/where_operators/Between.java
```diff
@@ -1,22 +1,27 @@
...
-  public V getMin() {
-    return getParameters().get(0);
-  }
+  public Between(V value1, V value2) {
+    super("BETWEEN ? AND ?");
 
-  public V getMax() {
-    return getParameters().get(1);
+    // this operator takes care of its own construction
+    if (!(value1 instanceof Column)) {
```
bpodgursky: would it work to have a separate constructor for Column arguments?  ex

public Between(Column min, Column max) 

it might be nicer than having the instanceof checks.

bpodgursky:
This looks great to me.  Whenever you're ready to merge, let me know and I can fix the workflow ui RldbQuery usages.  

The Record thing is no problem -- I have a hacky implementation of it in JackUtil.getModel(Class<T> type, Record fetch) which works fine for now.  Whenever we figure out a way to integrate it cleanly we can move it into Jack proper.

tuliren:
Thanks. I plan to merge tomorrow morning. There are a few other use cases already, so I can take care of the changes.

src/java/com/rapleaf/jack/queries/where_operators/Between.java
tuliren: Fixed. Created constructors and methods with different signatures to get rid of the instanceof check.

PR #120, opened 2015-04-17
Disallow calling find() on queries that call orderBy

SLieve:
@pwestling


SLieve:
@tuliren

src/rb/templates/i_query_builder.erb
```diff
@@ -0,0 +1,45 @@
...
+  <%= model_defn.query_builder_name %> <%= field_defn.name.camelcase(:lower) %>(<%= field_defn.java_type(true) %> value);
+
+  <%= model_defn.query_builder_name %> where<%= field_defn.name.camelcase() %>(IWhereOperator<<%= field_defn.java_type(true) %>> operator);
+
+  <%= model_defn.i_ordered_query_builder_name %> orderBy<%= field_defn.name.camelcase() %>();
+
+  <%= model_defn.i_ordered_query_builder_name %> orderBy<%= field_defn.name.camelcase() %>(QueryOrder queryOrder);
+
+  <% end %>
+}
```
tuliren: `i_query_builder.erb` and `i_ordered_query_builder.erb` need an extra newline at the end. Here is the reference: http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline

tuliren:
I think this design would allow the user to call `findWithOrder()` on an unordered query, because a `QueryBuilder` extends the `AbstractQueryBuilder`, which implements bothe the `IOrderedQueryBuilder` and the `IUnorderedQueryBuilder`. Will this be a concern?

There are a few test failures. For example:
```
java.lang.ClassCastException: com.rapleaf.jack.test_project.database_1.query.UserQueryBuilder cannot be cast to com.rapleaf.jack.test_project.database_1.query.IOrderedUserQueryBuilder
	at com.rapleaf.jack.test_project.database_1.query.UserQueryBuilder.order(UserQueryBuilder.java:61)
	at com.rapleaf.jack.TestModelQuery.testQueryWithOrder(TestModelQuery.java:245)
```

I have one concern that this design will introduce a lot of generated code for each model. An alternative is that the `OrderedQueryBuilder` only implements methods related to orders. So after the user calls any order method, the only thing she / he can do next is to call `findWithOrder`. The con is that the order clause can only be specified in the end. The pro is that less amount of generated code is needed. This is not necessarily a better solution, just something for discussion.

pwestling:
I don't think the amount of generated code is significant enough to worry about just yet. We should focus on ease-of-use first.

@SLieve Can we get the test errors and the formatting fixed up ASAP? I'd like to merge this before we forget about it :P

bpodgursky:
We should pay some attention to the amount of generated code, our rldb jar is already 20mb, and if we're talking doubling that it will actively hurt the jobtracker.

> The con is that the order clause can only be specified in the end. 

unless I'm misunderstanding, that's not really a big con, is it?  it's just a formatting thing.  that seems like a good idea to me.

pwestling:
I don't have any reason to believe that this would come anywhere close to doubling the amount of generated code. If it is, then yes that's a problem.

It's not a huge con, but it's enough of a usability hiccup, in my mind, to increase the generated code by 1% which is what I think we're talking about with stripping those methods off of the 1 generated interface.

The List-everywhere approach is obviously superior though, so I guess we can just drop this point and go with that.

bpodgursky:
yeah, that's fair.  not saying this particular change is a problem in code size, just that it is something to keep in mind with overall jack changes.

PR #121, opened 2015-04-21
Change interfaces and templates so that calling find() always returns a list

SLieve:
@bpodgursky @pwestling


SLieve:
@tuliren This is a PR for the other approach where find() returns a List and findWithOrder() is removed. Mind having a look?

test/regen_code.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-bundle exec ruby src/rb/jack.rb test/test_project/project.yml src/java
+bundle exec ruby src/rb/jack.rb test/test_project/project.yml test/java
```
tuliren: We are moving the test project to `test` directory? If that's the case, we should remove the old test project code in the `src`. Otherwise, they will be duplicated.

tuliren:
Looks good except that one comment.

PR #122, opened 2015-04-27
add missing associations

bpodgursky:
@pwestling seem ok?  two constructors were not setting associations.


pwestling:
Looks ok.Part of me wants to say that this constructor should not be used if IDatabases is null (and that silently not setting associations is a weird response), but I imagine we have tons of code that depends on being able to pass null.

bpodgursky:
I'm tempted to say that not much does since even passing the Attributes in to create the model is kind of an obscure use-case, and the models I've checked don't use it.  I'd generate db schemas locally and see what breaks, but something is borked in my gem config and I can't install jack.  I'll put up a sweep ticket to remove it.

PR #124, opened 2015-05-19
Lazily load each db in DatabasesImpl

SLieve:
@bpodgursky @tuliren 
This loads each db lazily instead of at construction time, which makes it possible for projects to only include required databases in the database.yml file.

### DIFF
--- a/test/java/com/rapleaf/jack/test_project/DatabasesImpl.java
+++ b/test/java/com/rapleaf/jack/test_project/DatabasesImpl.java
@@ -12,20 +12,18 @@ import com.rapleaf.jack.test_project.database_1.IDatabase1;
 import com.rapleaf.jack.test_project.database_1.impl.Database1Impl;
 
 public class DatabasesImpl implements IDatabases {
-  private final BaseDatabaseConnection database1_connection;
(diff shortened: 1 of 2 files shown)


test/java/com/rapleaf/jack/test_project/DatabasesImpl.java
```diff
@@ -12,17 +12,21 @@
...
+  private final BaseDatabaseConnection database1_connection;
+  private IDatabase1 database1;
 
   public DatabasesImpl(BaseDatabaseConnection database1_connection) {
-    this.database1 = new Database1Impl(database1_connection, this);
+    this.database1_connection = database1_connection;
   }
 
   public DatabasesImpl() {
-    this.database1 = new Database1Impl(new DatabaseConnection("database1"), this);
```
bpodgursky: I don't think this will do what we want though, because new DatabaseConnection("database1") will try to load the db's properties from the yaml.  We need that connection to be lazily instantiated.

SLieve: Oh derp, didn't know that the connection loaded the properties.

PR #126, opened 2015-05-29
Add support for getting connection from each database

SLieve:
@tuliren

### DIFF
--- a/src/rb/templates/db_impl.erb
+++ b/src/rb/templates/db_impl.erb
@@ -98,4 +98,8 @@ public class <%=db_name%>Impl implements I<%=db_name%> {
     return databases;
   }
 
+  public BaseDatabaseConnection getConnection() {
+    return conn;
+  }
+
 }
--- a/test/java/com/rapleaf/jack/test_project/database_1/impl/Database1Impl.java
+++ b/test/java/com/rapleaf/jack/test_project/database_1/impl/Database1Impl.java
@@ -100,4 +100,8 @@ public class Database1Impl implements IDatabase1 {
     return databases;
   }
 
+  public BaseDatabaseConnection getConnection() {
+    return conn;
+  }
+
 }


seancarr:
@SLieve what is this intended for? Previously this was intentionally unavailable.

SLieve:
@tuliren requested it, since he was using the connection for operations that jack doesn't support. With the multiple databases change he could no longer do:

    new DatabasesImpl(connection1);

since multiple connections are now required.

tuliren:
@seancarr The use case for this one is that we have a workflow that needs to do DELETE operation, so it needs a connection. Meanwhile it also use `rldb` for regular queries. Previously we constructed them like this:
```java
this.dbCon = new DatabaseConnection("rl_db");
this.rlDb = new DatabasesImpl(dbCon).getRlDb();
```

With the multiple db change, the construction would become
```java
this.dbCon = new DatabaseConnection("rl_db");
this.rlDb = new DatabasesImpl(dbCon, null).getRlDb();
```
But this approach is not ideal as new dbs are added, we need to add a new `null` each time.

So the current way is:
```java
this.dbCon = new DatabaseConnection("rl_db");
this.rlDb = new DatabasesImpl().getRlDb();
```

This approach constructs two connections. So if we can get the connection out of `rldb`, we can avoid having two connections while keep the constructor simple.

seancarr:
If the specific use case is to do deletes why not have jack support deletes? ActiveRecord does.
I think the main reason to not expose the connection to the consumer is to keep the interface simple and only expose things we want to allow the consumer to do, such as start/stop transactions, etc. If we expose the connection that would allow people to totally subvert jack and create their own prepared statements and stuff. I think that makes jack harder to maintain.

tuliren:
I see. I will leave the use case as it is and wait until Jack supports DELETE.

@SLieve Sorry for the extra work.

PR #129, opened 2015-07-16
Delete statement support

databanana:
@tuliren @bpodgursky: This is to add support for building DELETE statements from generated models.

### DIFF
--- a/src/java/com/rapleaf/jack/queries/WhereClause.java
+++ b/src/java/com/rapleaf/jack/queries/WhereClause.java
@@ -1,0 +1,96 @@
+package com.rapleaf.jack.queries;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Sets;
+
+import java.util.*;
+
+public class WhereClause {
+  List<WhereConstraint> whereConstraints;
+  Optional<Set<Long>> selectedIds;
+
+  public WhereClause() {
+    this.whereConstraints = new ArrayList<>();
+    this.selectedIds = Optional.absent();
+  }
+
+  public List<WhereConstraint> getWhereConstraints() {
+    return whereConstraints;
+  }
+
+  public Optional<Set<Long>> getIdSet() {
+    return selectedIds;
+  }
+
+  public void addConstraint(WhereConstraint constraint) {
+    whereConstraints.add(constraint);
+  }
+
+  public void addIds(Set<Long> ids) {
+    if (!selectedIds.isPresent()) {
+      selectedIds = Optional.<Set<Long>>of(Sets.<Long>newHashSet());
+    }
+    this.selectedIds.get().addAll(ids);
+  }
+
+  public void addId(Long id) {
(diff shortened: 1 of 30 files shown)


tuliren:
@databanana Thanks for the pull request.

One concern I have is that the delete builder is separate from the query builder, while it may be more elegant to share the builders between the query and the delete, and use different methods to trigger different operations on the built SQL statement like this:
```
users.query().whereNumPosts(greaterThan(3)).find();  // returns the query results
users.query().whereNumPosts(greaterThan(3)).delete();  // delete the entries
```

databanana:
@tuliren I originally thought that would be the best solution, but after thinking about the issue more my concerns were that SELECT statements have much broader functionality (i.e. GROUP BY, ORDER BY, and LIMIT clauses)--the only real commonality is that they both support a WHERE clause. This could lead to some confusing behavior, i.e. `users.query().whereNumPosts(greaterThan(3)).orderByNumPosts(QueryOrder.DESC).limit(3,4).delete()` would be valid syntax but wouldn't be supported in a DELETE statement. The QueryBuilder could throw an exception but it seems like it would be nice to enforce this restriction at compile time.

tuliren:
You're right. The syntax should be enforced at compile time. The delete operation does not support `select` and `group by` clauses (it does support `order` and `limit` though). So a separated builder is needed.

databanana:
I believe MySQL supports `order` and `limit` (but not `offset`), but my understanding is that this isn't standard SQL/PostgreSQL doesn't support this (correct me if I'm wrong, I'm not incredibly familiar with PostgreSQL) which is why I didn't add this to the delete builder.

src/java/com/rapleaf/jack/queries/WhereClause.java
```diff
@@ -0,0 +1,101 @@
+package com.rapleaf.jack.queries;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Sets;
+
+import java.util.*;
+
+public class WhereClause {
+  List<WhereConstraint> whereConstraints;
+  Optional<Set<Long>> selectedIds;
```
rfaugeroux: Could you get rid of the Optional here? It was not super useful in the first place, and since you're refactoring the `WhereClause` we might as well remove it now

rfaugeroux: Nevermind, I actually added that on purpose to fix a bug 6 months ago.

tuliren:
@databanana Sorry about the delay. All looks good, except that can you use `JackMatchers` in the test instead of constructing new operators? E.g. replace `new In<>(5, 10)` with `JackMatchers.in(5, 10)` or `in(5, 10)` with static imports.

databanana:
@tuliren No worries. Sure, I made that change to the test.

tuliren:
@databanana Thanks.

@seancarr Romain and I have reviewed the PR. All looks good. Can we merge?

PR #130, opened 2015-07-27
Add .ruby-version for newer versions of ruby

SLieve:
@tuliren @seancarr I had to do

    rbenv local 1.9.3-p551

Before generating because it was giving me a version mismatch error. This would prevent that in the future.


seancarr:
.ruby-version is specific to rbenv which jack doesn't rely on so I'd rather not add this file and instead have listed in the readme/dependencies specific versions of ruby that are supported.

@SLieve, what was the error? Were you using 1.8.7 or 2.x?

SLieve:
I see, makes sense. This was the error:

    rake aborted!
    dlopen(/Users/slieve/.rbenv/versions/ree-1.8.7-2011.03/lib/ruby/1.8/i686-darwin13.3.0/openssl.bundle, 9): Symbol not found: _SSLv2_client_method
      Referenced from: /Users/slieve/.rbenv/versions/ree-1.8.7-2011.03/lib/ruby/1.8/i686-darwin13.3.0/openssl.bundle
      Expected in: flat namespace
     in /Users/slieve/.rbenv/versions/ree-1.8.7-2011.03/lib/ruby/1.8/i686-darwin13.3.0/openssl.bundle - /Users/slieve/.rbenv/versions/ree-1.8.7-2011.03/lib/ruby/1.8/i686-darwin13.3.0/openssl.bundle
    /Users/slieve/code/jack/test/test_project/database_1/config/environment.rb:5

I was using 1.8.7

seancarr:
Ok, I think that error has to do with open-ssl not being properly setup for your ruby 1.8.7, not that ruby 1.8.7 doesn't work for jack. I think no action is required for now.

PR #131, opened 2015-07-27
Lazily load tables

SLieve:
@tuliren I think this should work for lazily loading tables.

### DIFF
--- a/src/rb/templates/db_impl.erb
+++ b/src/rb/templates/db_impl.erb
@@ -34,15 +34,12 @@ public class <%=db_name%>Impl implements I<%=db_name%> {
   private final IDatabases databases;
 
 <% model_defns.each do |model_defn| %>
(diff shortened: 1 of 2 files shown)


seancarr:
@SLieve, do you have an example of what type of error/exception this is solving for? That will help with the review.

tuliren:
@SLieve I think you also need to make the static `TABLE` and `COLUMN` variables lazy-loading. Otherwise they will cause trouble during the db schema migration as well. This is also an opportunity to correct the naming convention (i.e., resolving the mistake I made that leads to compilation error in the generated code if the user has a column named `TBL`).

tuliren:
Talked with @SLieve offline. The above statement is incorrect. `TABLE`s and `COLUMN`s won't affect production during the schema change.

SLieve:
Upon further inspection of what Jack does when first loading tables, this PR won't solve what it was trying to solve.

PR #132, opened 2015-09-03
include id in toString for jack models

LaneRamp:


seancarr:
@LaneRamp did you run the jack tests locally? I think this PR should include some updates to the generated model code.

LaneRamp:
local tests all passed for me without problem

bpodgursky:
could you check in the changes to the generated test fixture models?

test/java/com/rapleaf/jack/TestModelWithID.java
```diff
@@ -33,6 +34,9 @@ public void setUp() {
 
   public void testFields() {
 
+    assertTrue(postModel.toString().equals("<Post id: 0 title: Test Post posted_at_millis: 100 user_id: 1 updated_at: 0>"));
```
bpodgursky: for regression testing purposes, can you have a test that verifies the ID is something real after it's saved?  you can re-fetch the model after save(userModel); is called and it should have a real ID

test/java/com/rapleaf/jack/test_project/database_1/models/Comment.java
```diff
@@ -33,7 +33,7 @@
 
 public class Comment extends ModelWithId<Comment, IDatabases> implements Comparable<Comment>{
   
-  public static final long serialVersionUID = 8036617193786054443L;
+  public static final long serialVersionUID = 6213989608937906012L;
```
seancarr: I'm guessing this change is due to this being generated with ruby 1.9.3. Currently this project is still expecting ree 1.8.7. Can you generate with that and see if that keeps the ids the same?

PR #133, opened 2015-09-04
Support where clauses for ids

michel-tricot:
Make it possible to do:
```
MyModel.query().whereId(myWhereClause).find()
```

Today it is only supported for non-id fields

### DIFF
--- a/src/java/com/rapleaf/jack/queries/WhereConstraint.java
+++ b/src/java/com/rapleaf/jack/queries/WhereConstraint.java
@@ -36,6 +36,10 @@ public class WhereConstraint<T> implements QueryCondition {
     return column.getField();
   }
 
+  public boolean isId() {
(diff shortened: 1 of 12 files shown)


michel-tricot:
Code compiles, I still need to add some ```whereId``` tests. 

Unfortunately ```mvn test``` doesn't seem to work out of the box so I will have to play a little with my environment.

michel-tricot:
A follow up task is to remove the hardcoded ```idIn``` and all the code associated with it.

michel-tricot:
Should be all good

bpodgursky:
@tuliren how does this look to you?  looks good to me.

tuliren:
All looks good!

michel-tricot:
@bpodgursky can you merge this PR?

PR #134, opened 2015-09-06
Lazy load table metadata

michel-tricot:
This change will make it easier to delete table from the DB so that you don't have to synchronize the migration with the java code deploy.

The code will only fail if you try to *use* a table which doesn't exist in the db instead of failing on Database instantiation.

### DIFF
--- a/src/java/com/rapleaf/jack/LazyLoadPersistence.java
+++ b/src/java/com/rapleaf/jack/LazyLoadPersistence.java
@@ -1,0 +1,39 @@
+package com.rapleaf.jack;
+
+public abstract class LazyLoadPersistence<T extends IModelPersistence, D extends GenericDatabases> {
+
+  private final BaseDatabaseConnection conn;
+  private final D databases;
+
+  private volatile T persistence;
+
+  private boolean disableCaching;
+
+  public LazyLoadPersistence(BaseDatabaseConnection conn, D databases) {
+    this.conn = conn;
+    this.databases = databases;
+
+    this.persistence = null;
+    this.disableCaching = false;
+  }
+
+  public T get() {
+    if (persistence == null) {
+      synchronized (this) {
+        this.persistence = build(conn, databases);
+      }
+    }
+
+    if (disableCaching) {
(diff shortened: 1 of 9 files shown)


src/java/com/rapleaf/jack/LazyLoadPersistence.java
```diff
@@ -0,0 +1,42 @@
...
+    this.conn = conn;
+    this.databases = databases;
+
+    this.persistence = null;
+    this.disableCaching = false;
+  }
+
+  public T get() {
+    if (persistence == null) {
+      synchronized (this) {
```
bpodgursky: is the synchronization around trying to prevent people from simultaneously using the database connection?  if so, I think there's a case where you call get() on two persistences simultaneously and it'll use it at the same time.  if that's the goal then synchronizing on conn seems safer.

bpodgursky:
one comment, seems good otherwise.

tuliren:
@SLieve, is this pull request similar to yours? I remember that last time you said your approach would not resolve the problem due to the way Jack loads the tables. Do you think this pull request bypass the issue?

src/java/com/rapleaf/jack/LazyLoadPersistence.java
michel-tricot: That was not the goal. Should I be worried about that case?

The goal here was to make sure that we don't create two instances if the same persistence since it will be created at usage and not at init time like before.

bpodgursky: I think to prevent two persistences from being created you'd have to synchronize outside the if statement, right?

SLieve:
Yes, this is pretty much what I wanted to do. Looks good to me.

src/java/com/rapleaf/jack/LazyLoadPersistence.java
michel-tricot: Actually I forgot an if in the synchronize block. Just fixed it. 

I need the two ifs.

PR #136, opened 2015-09-17
Allowing disabling then re-enabling caching of lazily loaded persistences

databanana:
@tuliren @bpodgursky @michel-tricot: I added some test cases for and fixed an issue in which once caching is disabled on a lazily loaded model persistence, it cannot be re-enabled.

### DIFF
--- a/test/java/com/rapleaf/jack/BaseDatabaseModelTestCase.java
+++ b/test/java/com/rapleaf/jack/BaseDatabaseModelTestCase.java
@@ -13,6 +13,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.rapleaf.jack.test_project.database_1.IDatabase1;
 import junit.framework.TestCase;
 
 import com.rapleaf.jack.test_project.IDatabases;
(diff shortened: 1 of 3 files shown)


test/java/com/rapleaf/jack/BaseDatabaseModelTestCase.java
```diff
@@ -598,4 +599,14 @@ public void testComparable() {
     assertEquals(post3, posts.get(2));
     assertEquals(post4, posts.get(3));
   }
+
+  public void testDisableThenEnableCaching() {
```
databanana: I don't completely understand the separation between `BaseDatabaseModelTest` and `TestAbstractDatabaseModel`–is this the right place to put this test?

bpodgursky:
looks good to me.  @michel-tricot this seem right?

test/java/com/rapleaf/jack/BaseDatabaseModelTestCase.java
bpodgursky: yeah... I'm not sure either.  this seems fine.

PR #140, opened 2015-11-30
Retrieve Model and Attributes from Record

tuliren:
@bpodgursky, this pull request incorporates your `JackUtil` into the `Record` and `Records` classes. The main change is that the new method takes one `Table` parameter, instead of the model and attributes types. Those information is wrapped in `Table`. So the method call is slightly simplier. A bug fix is also included in this pull request. It turns out that SQL sometimes returns double value as `BigDecimal` or `Float`. These types are put in special checks.

Since `Attributes` is already a plural word, the methods for retrieving attributes are the same for `Record` and `Records`: `Record#getAttributes` returns one `Attributes`, and `Records#getAttributes` returns a list of it.

If the user tries to retrieve the model / attributes of a table whose fields do not exist in the `Record`, `null` will be returned. I think `null` is legitimate since outer join is supported in the generic query. However, I am not sure if it is better to return an `Optional` and use `Optional.absent()` to replace the `null` case. If there is no objection, I will keep returning `null`.

I still have issues with the `UID`. When I use ruby `ree-1.8.7`, the `UID` changes, and there is one more change in `Gemfile.lock`:
```
 DEPENDENCIES
   activesupport (>= 3.0.0)
-  i18n
+  i18n (= 0.5.0)
```
When I use ruby `1.9.3`, the `UID` remains the same, and there is again a change in `Gemfile.lock`:
```
 DEPENDENCIES
   activesupport (>= 3.0.0)
-  i18n
+  i18n (= 0.5.0)
+
+BUNDLED WITH
+   1.10.5
```
It looks like the change in `Gemfile` is inevitable. I suspect that someone used `1.9.3` in a previous pull request and changed the `UID` accidentally. So I should probably push the `UID` change with `ree-1.8.7` this time to revert them back. Will check with `Admin`.

We should probably add a `.ruby-version` file to prevent the confusing.

### DIFF
--- a/src/java/com/rapleaf/jack/queries/Record.java
+++ b/src/java/com/rapleaf/jack/queries/Record.java
@@ -1,6 +1,7 @@
 package com.rapleaf.jack.queries;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.math.BigDecimal;
 import java.util.Date;
 import java.util.Map;
@@ -8,6 +9,7 @@ import java.util.Map;
 import com.google.common.collect.Maps;
 
 import com.rapleaf.jack.AttributesWithId;
+import com.rapleaf.jack.GenericDatabases;
 import com.rapleaf.jack.ModelWithId;
 import com.rapleaf.jack.util.JackUtility;
 
@@ -124,8 +126,16 @@ public class Record {
     return attribute;
   }
 
-  public <M extends ModelWithId> M getModel(Class<M> modelType) {
-    return null;
+  @SuppressWarnings("unchecked")
+  public <M extends ModelWithId, D extends GenericDatabases> M getModel(Table tableType, D databases) {
+    try {
+      Constructor<M> constructor = (Constructor<M>)(tableType.getModelType().getConstructor(tableType.getAttributeType(), databases.getClass().getInterfaces()[0]));
+      M model = constructor.newInstance(getAttributes(tableType), databases);
+      model.setCreated(true);
+      return model;
+    } catch (NoSuchMethodException | InvocationTargetException | IllegalAccessException | InstantiationException e) {
+      throw new RuntimeException(e);
+    }
   }
 
   private Object checkTypeAndReturnObject(Column column, Class clazz) {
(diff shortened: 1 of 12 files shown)


src/java/com/rapleaf/jack/queries/Record.java
```diff
@@ -72,6 +77,74 @@ public Boolean getBoolean(Column column) {
...
+    Map<Enum, Object> fieldMap = Maps.newHashMap();
+    for (Map.Entry<Column, Object> entry : columns.entrySet()) {
+      Column column = entry.getKey();
+      if (column.getTable().equals(tableName)) {
+        if (column.getField() != null) {
+          fieldMap.put(column.getField(), entry.getValue());
+        } else if (id == null) {
+          id = ((Number)(entry.getValue())).longValue();
+        } else {
+          throw new RuntimeException("The record contains multiple IDs for Table " + tableName);
```
bpodgursky: so I'm not sure how fixable it is, but this was a bit of a hack at the time.  column.getField() is null for ID fields, so I guess that it's the ID.  but the problem is that if your Record has columns from multiple tables, it can't figure out which ID is which.  it could also get the entirely wrong ID if you selectAttributes for a record type which doesn't correspond to the ID present.

I think this is still useful to have whether or not it's fixed, just FYI in case you see an easy fix.

bpodgursky:
Yeah, I'm not really sure what the state of ruby is here.  I don't think the uuid in the generated classes here should matter much, as long as it doesn't change in downstream projects we generate with jack.  So that seems reasonable.

src/java/com/rapleaf/jack/queries/Record.java
tuliren: This problem is actually fixed in this pull request. This check, `column.getTable().equals(tableName)`, filters the fields by the table name to prevent it. In the test case, I did test it with a join query from two tables:
```
Record record = db.createQuery()
    .from(User.TBL)
    .innerJoin(Post.TBL).on(Post.USER_ID.equalTo(User.ID))
    ...
```

The `else` branch could only happen if someone manually constructs a `Record` with two `ID` fields from the same table, which is highly unlikely.

bpodgursky: got it.  missed that check.  nice.

PR #141, opened 2015-12-03
[DT-93] Type Safe Generic Query

tuliren:
This pull request incorporates type safety into the generic query. This is achieved by storing the field type in the column object.
```java
// Column for id field
public static Column<Long> fromId(String table) {
  return new Column<Long>(table, null, Long.class);
}
// Column for datetime field
public static Column<Long> fromTimestamp(String table, Enum field) {
  return new Column<Long>(table, field, java.sql.Timestamp.class);
}
// Column for date field
public static Column<Long> fromDate(String table, Enum field) {
  return new Column<Long>(table, field, java.sql.Date.class);
}
// All other columns
public static <T> Column<T> fromField(String table, Enum field, Class<T> fieldType) {
  return new Column<T>(table, field, fieldType);
}
```
This information is then used for improved query construction and result retrieval.

### Type-Safe Query
Each column is typed and can only be queried with parameters or other columns of the same type.
```java
// Compilable
User.SOME_FLOAT.equalTo(1.0)
User.SOME_FLOAT.lessThan(User.SOME_DECIMAL)

// Compile-time error
User.SOME_FLOAT.equalTo(1L)
User.SOME_FLOAT.lessThan(User.SOME_DATE)
```

If a query is needed between two columns of different types, an explicit type conversion is required.
```java
User.ID.equalTo(Comment.COMMENTER_ID.as(Long.class))
```

### One Method to Get Them All
The new `Record#get` and `Records#gets` methods allow users to get the value for a field with the correct type.
```java
// Previous methods for field value retrieval
Long userId = record.getLong(User.ID);
String bio = record.getString(User.BIO);
List<Double> someDecimals = records.getDoubles(User.SOME_DECIMAL);

// New methods for field value retrieval
Long userId = record.get(User.ID);
String bio = record.get(User.BIO);
List<Double> someDecimals = records.gets(User.SOME_DECIMAL);
```

The `get<type>` and `get<type>s` methods are no longer needed. Maybe they should be deprecated in a following pull request.

### Auto Creation of Date and Datetime Literal
Previously, to specify a where clause with `Date` or `Datetime` columns, `DATE` and `DATETIME` functions must be used to convert a `long` parameter to a date literal.
```java
// Previous incorrect date query
where(User.SOME_DATETIME.equalTo(System.currentTimeMillis()))
// Previous correct date query
where(User.SOME_DATETIME.equalTo(DATETIME(System.currentTimeMillis())))
```
These functions can be easily forgotten, resulting in an empty query, which has caused lots of frustrations.

Now with the typed columns, the date functions are no longer needed.
```java
// Current correct date query
where(User.SOME_DATETIME.equalTo(System.currentTimeMillis()))
where(User.SOME_DATETIME.equalTo(DATETIME(System.currentTimeMillis()))) // deprecated
```

These functions are exempted from deletion in this pull request for backward compatibility. They just pass through the input values and performs no operation at all. They will be removed in a following pull request, once we migrate all production code to the new Jack.

### Fixed Aggregation Function
Previously, the return type of an aggregated field is always the same as the original field. This is expected for `MAX`, `MIN`, and `SUM`, but incorrect for `COUNT` and `AVG`. The return type of `AVG` has been changed to `Number`, so that the users can convert it to suitable type at will.
```java
Integer intAverage = record.getNumber(AVG(User.NUM_POSTS)).intValue();
Double doubleAverage = record.getNumber(AVG(User.NUM_POSTS)).doubleValue();
```

The return type of `COUNT` still follows the type of the original field. It will be changed to `Integer` once this pull request has been merged and the production code has been updated with the new Jack.

### Plan for Deployment
1. Merge the pull request. Most changes should be backward compatible.
2. Update production use cases to the new code, including:
  - Replace `get<type>` and `get<type>s` with `get` and `gets` methods;
  - Remove date and datetime functions;
  - Add column type conversion for operations between columns of different types.
3. Create new pull request to remove the date and datetime `Functions`, as well as change the return type of `COUNT`.


test/java/com/rapleaf/jack/TestGenericQuery.java
```diff
@@ -153,111 +144,7 @@ public void testBasicQuery() throws Exception {
             .or(User.BIO.equalTo("Trader"), User.HANDLE.equalTo("A")))
         .fetch();
     assertEquals(3, results1.size());
-    assertEquals(Sets.newHashSet(userA.getId(), userB.getId(), userC.getId()), Sets.newHashSet(results1.getLongs(User.ID)));
-  }
-
-  @Test
```
bpodgursky: do these tests don't not make sense anymore?

bpodgursky:
this is awesome.  looks great.

test/java/com/rapleaf/jack/TestGenericQuery.java
tuliren: The `TestGenericQuery` file is very large. So I broke the code into two files, one just for the query logic, and one for `Record` and `Records`. These tests have been moved to `TestRecordAndRecords `.

tuliren:
I have not done a thorough analysis of the impact of this change upon our production code. Probably won't have time to finish that until Sunday or next Monday.

@seancarr, could you do a second round of review?

seancarr:
@tuliren I'm pretty swamped and don't know that I'd actually be able to give it a thorough look within the next week. Do you think it's important for me to review it?

tuliren:
@seancarr, got it. Don't worry about it then. I will try to do a thorough production check before merging.

PR #146, opened 2016-03-30
Alpha Redshift support in Jack

michel-tricot:
The way ids are handled in jack make this change dirty. It always assumes that `id` is an existing column based on Rails convention. Unfortunately there is no such assumption in redshift and supporting this  requires a huge refactoring.

Discovered and fixed a potential bug in code generation: if a column in the table is named `tbl` it will generate invalid code. The fix is just to prefix the guilty column with an `_`. A proper fix would require a change in how we namespace the different fields in the jack models and this change is not backward compatible.


michel-tricot:
@tuliren @seancarr Can you take a look at this PR?

src/rb/templates/persistence_impl.erb
```diff
@@ -209,7 +209,12 @@ public class <%= model_defn.impl_name %> extends AbstractDatabaseModel<<%= model
   @Override
   protected <%= model_defn.model_name %> instanceFromResultSet(ResultSet rs, Set<Enum> selectedFields) throws SQLException {
     boolean allFields = selectedFields == null || selectedFields.isEmpty();
-    return new <%= model_defn.model_name %>(rs.getLong("id"),
+    <% if database_defn.disable_rails_id %>
+    long id = -1;
```
seancarr: Will this mess with the caching by id that happens in the persistences?

michel-tricot: Yes it will but only if you activate the "quick and dirty" option.
I will disable persistence by default in that case

src/java/com/rapleaf/jack/queries/WhereConstraint.java
```diff
@@ -54,6 +54,6 @@ public Logic getLogic() {
 
   @Override
   public String getSqlStatement() {
-    return (logic != null ? logic.toString() + " " : "") + "`" + column.getSqlKeyword() + "` " + operator.getSqlStatement();
```
tuliren: The backticks are necessary to prevent keyword collisions (PR #138). E.g. if there is a column named `group`, the where clause for this column needs to wrap `group` in these backticks to prevent `MySQLSyntaxErrorException`.

tuliren:
Looks good. Just that one comment.

src/java/com/rapleaf/jack/queries/WhereConstraint.java
michel-tricot: Hum this is annoying it is only supported in mysql

tuliren: Ansi standard uses double quotes for delimited identifiers, which is also supported by postgresql and redshift. However, to use double quotes in mysql, we need to specify the ansi mode: [reference](http://dev.mysql.com/doc/refman/5.0/en/identifiers.html). So it seems that we have to make the query aware of the database type.

michel-tricot: ok, I am adding it back and I am making my change dirtier.
I will create a seperate PRs to support backticks and double quotes

src/java/com/rapleaf/jack/DatabaseConnection.java
```diff
@@ -101,16 +109,32 @@ public Connection getConnection() {
...
       updateExpiration();
       return conn;
-    } catch(Exception e) { //IOEx., SQLEx.
+    } catch (Exception e) { //IOEx., SQLEx.
       throw new RuntimeException(e);
     }
   }
 
+  @Override
+  public PreparedStatement getPreparedStatement(String statement) {
```
michel-tricot: @tuliren FYI

tuliren: Looks good.

PR #149, opened 2016-06-18
Move createQuery up to IDb interface

shreyaskumar:
This allows us to write code that operates on any DB. Our use case is generic status setting across models.


bpodgursky:
looks reasonable to me.  @tuliren ?

tuliren:
:shipit:

PR #150, opened 2016-06-20
Add ability to specify update time for testing

TV4Fun:
For tests that rely on updating the database, it would be extremely helpful
to be able to set an update time for the updated_at column to be set to,
rather than having this hardcoded into the save method. This adds that
while also keeping the default behavior.

@bpodgursky

### DIFF
--- a/src/java/com/rapleaf/jack/AbstractDatabaseModel.java
+++ b/src/java/com/rapleaf/jack/AbstractDatabaseModel.java
@@ -622,8 +622,8 @@ public abstract class AbstractDatabaseModel<T extends ModelWithId<T, ? extends G
     return (Integer)model.getField(lockFieldName);
   }
 
-  public boolean saveStrict(T model) throws JackException, IOException {
-    Long oldUpdatedAt = handleRailsUpdatedAt(model);
+  public boolean saveStrict(long updateTimeMillis, T model) throws JackException, IOException {
+    Long oldUpdatedAt = handleRailsUpdatedAt(updateTimeMillis, model);
     if (model.isCreated()) {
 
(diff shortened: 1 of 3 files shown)


bpodgursky:
I can't think of any reason not.  @tuliren?  

kinda related, looking through the code, I wonder if we should even be generating setUpdatedAt on the models, since it's not respected when saving.

roshan:
There's currently a bug where if `updatedAt` is not set, it will never be set.

bpodgursky:
@roshan it seems pretty straightforward to just remove the null check in updatedAtCanBeHandled, right? make it

  private boolean updatedAtCanBeHandled(T model) {
    if (model.hasField("updated_at")) {
      return (field.getClass().equals(Long.class));
    }
    return false;
  }

that code is super old though; @seancarr do you know why there's a null check here?  https://github.com/LiveRamp/jack/blob/master/src/java/com/rapleaf/jack/AbstractDatabaseModel.java#L846

roshan:
To be honest, I think that's pretty much it, but I don't remember why it's the other way so I didn't do it.

tuliren:
Sorry about the delayed reply. No internet yesterday.

Looks good. @TV4Fun, can you add a simple test? (Similar to `TestModelWithID#testUpdatedAt`).

TV4Fun:
Added a test called `testManualUpdatedAt` to `TestModelWithID`

bpodgursky:
@TV4Fun looks good to me.  I added you to the group so you should be able to merge.  when you do, could you commit the updated source for db_schemas and census_db?  so the builds don't complain about the diff.

PR #151, opened 2016-06-28
Remove redundant `public` specifiers from Jack-generated interfaces

TV4Fun:
The `public` specifier Jack puts on its interface methods is unnecessary and
clutters the output code. This change removes that.

@bpodgursky


bpodgursky:
looks good

PR #154, opened 2016-07-14
add support for namespaced models

sidoh:
@tuliren @bpodgursky -- Can you guys take a look? Tested this with rldb and workflow_db. Caused a minor change in rldb, but I think it's the correct behavior.

The intention here is to add support for namespaced models. We want to do this as we're adding new DBs to avoid naming conflicts.

This PR does two things:

1. Looks for models in recursive subdirs of models dir. Previously only looked for `.rb` files in `app/models`.
2. Strips namespaces from class names (e.g., `MyNamespace::Model` --> `Model`). This assumes that the table naming convention for namespaced models is `MyNamespace::Model` is `models` instead of `my_namespace_models`. I don't see a nice way around this without completely refactoring Jack to not use dumb regex parsing.

### DIFF
--- a/src/rb/models_dir_processor.rb
+++ b/src/rb/models_dir_processor.rb
@@ -1,11 +1,11 @@
 # Copyright 2011 Rapleaf
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #    http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


tuliren:
Looks good to me. What is the minor change in rldb? So the test dbs are not affected?

sidoh:
Minor change in rldb was due to some model in a nested directory having an association to a different model that wasn't. Since nested dirs weren't being considered before, the Jack association wasn't generated.

What's the best way to test the jack changes? Was doing:

```
$ bundle exec ruby src/rb/jack.rb test/test_project/project.yml test/java
```

But this seems to result in changes that shouldn't be the result of this PR. Removing `public` from interfaces and such. Perhaps we forgot to regenerate the test models in a previous PR?

bpodgursky:
looks reasonable.

> Perhaps we forgot to regenerate the test models in a previous PR?

plausible

tuliren:
Confirmed that `public` was removed in this https://github.com/LiveRamp/jack/pull/151, but the code was not regenerated.

sidoh:
Confirmed that there are no changes to the test project.

Maybe we should add something to the travis build to assert there aren't changes?

PR #156, opened 2016-07-28
Fix syntax error generation in database.yml

TV4Fun:
If update_database_yml.sh is called multiple times, it will leave
a syntax error in database.yml. This fixes that.

@bpodgursky


bpodgursky:
lol

PR #157, opened 2016-07-28
Support for ruby2

sidoh:
This adds support for ruby 1.9+ and rails 4+. It also fixes a bug where default values for dates were never respected.

I confirmed that:

* There were no significant changes to generated models in db_schemas or workflow_db for ruby 1.8/rails3.
* This works for ruby 2.3/rails 4. 
* There are no substantive changes to db_schemas or workflow_db when I upgrade them to ruby 2.3/rails4.

Unfortunately upgrading to rails 4 results in a bunch of serial version UID changes because rails 4 adds `length: <x>` to `schema.rb` even when `length` is not provided by a migration (this was not the case in rails3). Since serial version UID generation is affected by column arguments, this causes it to change. We'll probably just have to deal with this if we want to upgrade to rails 4.

(I apologize for all of the whitespace changes. Only realized my editor was adding them after I made a bunch of changes.)

@tuliren @bpodgursky can you guys take a look? think this is mostly ready to go.


tuliren:
:shipit:

sidoh:
Gonna merge this on Monday unless there are objections.

PR #159, opened 2016-08-01
Get rid of double curly brace initialization in persistence impls

TV4Fun:
Using double curly braces to initialize map and list objects is an anti-pattern
that we should never use. See https://blog.jooq.org/2014/12/08/dont-be-clever-the-double-curly-braces-anti-pattern/
for a detailed discussion. Essentially the problems are:
 * It's not very readable.
 * It creates a separate class for each instance, which adds overhead in the
   ClassLoader.
 * It can create memory leaks, since the new objects retain a reference to
   their enclosing class, which leaves a high potential for circular
   references.

This removes this anti-pattern from our jack-generated code.
@bpodgursky

### DIFF
--- a/src/rb/templates/persistence_impl.erb
+++ b/src/rb/templates/persistence_impl.erb
@@ -220,7 +220,9 @@ public class <%= model_defn.impl_name %> extends AbstractDatabaseModel<<%= model
   <% model_defn.fields.each do |field_defn| %>
 
   public List<<%= model_defn.model_name %>> findBy<%= field_defn.name.camelcase %>(final <%= field_defn.java_type %> value) throws IOException {
-    return find(new HashMap<Enum, Object>(){{put(<%= model_defn.model_name %>._Fields.<%= field_defn.name %>, value);}});
+    Map<Enum, Object> searchKey = new HashMap<>(1, 1.0f);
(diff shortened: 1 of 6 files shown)


src/rb/templates/persistence_impl.erb
```diff
@@ -220,7 +220,9 @@ public class <%= model_defn.impl_name %> extends AbstractDatabaseModel<<%= model
   <% model_defn.fields.each do |field_defn| %>
 
   public List<<%= model_defn.model_name %>> findBy<%= field_defn.name.camelcase %>(final <%= field_defn.java_type %> value) throws IOException {
-    return find(new HashMap<Enum, Object>(){{put(<%= model_defn.model_name %>._Fields.<%= field_defn.name %>, value);}});
+    Map<Enum, Object> searchKey = new HashMap<>(1, 1.0f);
```
bpodgursky: could we use Collections.singletonMap?  nbd, just a bit terser.

bpodgursky:
small comment, looks reasonable.

src/rb/templates/persistence_impl.erb
TV4Fun: Done.

bpodgursky:
:shipit:

PR #164, opened 2016-08-05
if the connection is already closed, don't refuse to reset

twshaw3:
@seancarr @npoulallion FYI

### DIFF
--- a/src/java/com/rapleaf/jack/BaseDatabaseConnection.java
+++ b/src/java/com/rapleaf/jack/BaseDatabaseConnection.java
@@ -39,7 +39,7 @@ public abstract class BaseDatabaseConnection implements Serializable {
     LOG.warn("Resetting database connection to attempt to recover from exception: {}", cause);
     if (conn != null) {
       try {
-        if (!conn.getAutoCommit()) {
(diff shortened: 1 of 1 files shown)


src/java/com/rapleaf/jack/BaseDatabaseConnection.java
```diff
@@ -39,7 +39,7 @@ public Connection resetConnection(Throwable cause) {
     LOG.warn("Resetting database connection to attempt to recover from exception: {}", cause);
     if (conn != null) {
       try {
-        if (!conn.getAutoCommit()) {
+        if (!conn.isClosed() && !conn.getAutoCommit()) {
```
seancarr: @twshaw3, I think this could have undesirable consequences. The new connection will not have auto commit set to false. And if it did it would potentially execute what would be the 2nd part of a transaction as it's own transaction.

Basically a connection being closed in the middle of a transaction is not a safe thing. If we could ensure we weren't in the middle of the transaction it would be okay but I don't think we can ensure that, particularly if the connection is closed.

twshaw3: Interesting, and makes sense. Going to close.

PR #165, opened 2016-08-06
Alpha Redshift support

michel-tricot:
Safer change than https://github.com/LiveRamp/jack/pull/146

The way ids are handled in jack make this change dirty. It always assumes that id is an existing column based on Rails convention. Unfortunately there is no such assumption in redshift and supporting this requires a huge refactoring.

Discovered and fixed a potential bug in code generation: discard columns named tbl and id as they may conflict with existing class members

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.amazon.redshift</groupId>
+      <artifactId>jdbc41</artifactId>
+      <version>1.1.10.1010</version>
+    </dependency>
+
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>13.0.1</version>
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
     <dependency>
       <groupId>com.amazon.redshift</groupId>
       <artifactId>jdbc41</artifactId>
-      <version>1.1.10.1010</version>
+      <version>1.1.17.1017</version>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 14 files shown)


michel-tricot:
@bpodgursky can you take quick look at it?

pom.xml
```diff
@@ -83,6 +83,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.amazon.redshift</groupId>
```
sidoh: I don't think this is in maven central. Amazon's awful about not putting their stuff in maven central.

Probably have to find a mirror or check in the jar with the project.

bpodgursky: we can put it libs-ext-release at repository.liveramp.com if it's not in maven central (we use it here).  checking in jars with the project is pretty awful.

src/rb/templates/persistence_impl.erb
```diff
@@ -209,7 +209,17 @@ public class <%= model_defn.impl_name %> extends AbstractDatabaseModel<<%= model
   @Override
   protected <%= model_defn.model_name %> instanceFromResultSet(ResultSet rs, Set<Enum> selectedFields) throws SQLException {
     boolean allFields = selectedFields == null || selectedFields.isEmpty();
-    return new <%= model_defn.model_name %>(rs.getLong("id"),
+    <% if database_defn.support_missing_id %>
+    long id;
+    try {
+      id = rs.getLong("id");
+    } catch (SQLException e) {
```
bpodgursky: catching this as an exception seems pretty ugly.  could we check for the column as described here?

http://stackoverflow.com/questions/3599861/how-can-i-determine-if-the-column-name-exist-in-the-resultset

pom.xml
michel-tricot: fixed

src/rb/templates/persistence_impl.erb
michel-tricot: fixed

PR #167, opened 2016-08-30
allow overriding configs via environment

pwestling:
Allow the setting or overriding of all Jack connection configurations by
preferring data from environment variables and system properties. This
allows the use of Jack in situations where a config file is hard to
supply, and allows overriding specific properties for non-standard use
cases.

@roshan and @sidoh, as the most recent contributor and as a dev tools guy, if you're willing to quickly review I'd appreciate it.

### DIFF
--- a/src/java/com/rapleaf/jack/DatabaseConnectionConfiguration.java
+++ b/src/java/com/rapleaf/jack/DatabaseConnectionConfiguration.java
@@ -16,10 +16,10 @@ public class DatabaseConnectionConfiguration {
   private String dbName;
   private Optional<Integer> port;
   private Optional<Boolean> parrallelTest;
-  private String username;
-  private String password;
+  private Optional<String> username;
+  private Optional<String> password;
 
-  public DatabaseConnectionConfiguration(String adapter, String host, String dbName, Optional<Integer> port, Optional<Boolean> parrallelTest, String username, String password) {
+  public DatabaseConnectionConfiguration(String adapter, String host, String dbName, Optional<Integer> port, Optional<Boolean> parrallelTest, Optional<String> username, Optional<String> password) {
     this.adapter = adapter;
     this.host = host;
     this.dbName = dbName;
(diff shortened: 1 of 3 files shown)


bpodgursky:
It's going to be pretty limiting if the environment variables aren't for a specific DB.  

It's kind of ugly, but looking for environment variables like JACK_DB_USERNAME_RL (for rl_db), or something along those lines would let you specify creds per DB.

pwestling:
Didn't even occur to me. Good point, thanks Ben. Will fix.

sidoh:
👍

src/java/com/rapleaf/jack/DatabaseConnectionConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+import com.google.common.collect.Maps;
+import org.jvyaml.YAML;
+
+public class DatabaseConnectionConfiguration {
+
+  private String adapter;
+  private String host;
+  private String dbName;
+  private Optional<Integer> port;
+  private Optional<Boolean> parrallelTest;
```
michel-tricot: typo, should be parallel

src/java/com/rapleaf/jack/DatabaseConnectionConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+    this.adapter = adapter;
+    this.host = host;
+    this.dbName = dbName;
+    this.port = port;
+    this.parrallelTest = parrallelTest;
+    this.username = username;
+    this.password = password;
+  }
+
+  public static DatabaseConnectionConfiguration loadFromEnvironment(String dbname_key) {
```
michel-tricot: Why snake case?

pwestling: That's what the var was in the calling class - I'll clean this up

PR #168, opened 2016-08-30
Specify ruby version to run on

TV4Fun:
This will help avoid the system selecting a default ruby version that Jack is incompatible with.

@bpodgursky


bpodgursky:
cool

PR #169, opened 2016-09-01
Update tests to use JUnit 4 style annotations

TV4Fun:
Replace the deprecated JUnit 3 style TestCase usage and update imports.

@bpodgursky


bpodgursky:
nice

PR #170, opened 2016-09-02
[DT-372] Index Hint

tuliren:
This pull request adds index hint to generic query.

The syntax is:
```java
Index index1 = Index.of("mock_index_1");
Index index2 = Index.of("mock_index_2");

db.createQuery()
    .from(User.TBL.with(IndexHints.use(index1), IndexHints.ignoreForOrderBy(index2)))
    .getSqlStatement();
```

Index hints can have different types: `USE`, `FORCE` and `IGNORE`, and different scoped: `ALL`, `JOIN`, `ORDER BY` and `GROUP BY`, according to MySQL 5.5 ([reference](http://dev.mysql.com/doc/refman/5.5/en/index-hints.html)).

For simplicity, the name of the index needs to be specified manually. Indices can be read from the schema and assigned to static variables in tables later on, but not in this pull request.

When calling `Table#use`, a `Table` is converted to a `TableReference`. This class will be expanded to support nested query in the future.

@bpodgursky, what do you think about the syntax?


bpodgursky:
this is awesome.  syntax looks straightforward to me.

tuliren:
According to [MySQL doc](http://dev.mysql.com/doc/refman/5.5/en/index-hints.html), `USE INDEX` AND `FORCE INDEX` cannot be mixed together. A runtime exception (`InvalidIndexHintException`) will thrown if that's the case. It is not a checked exception in that once it is thrown out, the exception cannot be recovered, so the user cannot do anything useful in the `catch` block. Therefore a runtime exception seems appropriate.

PR #172, opened 2016-10-05
changes for matching columns

dwarrier:
@rfaugeroux @tuliren 

I'm doing this because I would like to use a join that involves using a 'contains' condition against a column name instead of a string literal for another task.

### DIFF
--- a/src/java/com/rapleaf/jack/queries/where_operators/Match.java
+++ b/src/java/com/rapleaf/jack/queries/where_operators/Match.java
@@ -1,8 +1,16 @@
 package com.rapleaf.jack.queries.where_operators;
 
+import com.google.common.base.Preconditions;
+import com.rapleaf.jack.queries.Column;
+
 public class Match extends WhereOperator<String> {
 
   public Match(String pattern) {
     super("LIKE ?", pattern);
+  }
+
+  public Match(Column<String> column, String prefix, String suffix) {
+    super(String.format("LIKE CONCAT('%s', %s, '%s')", prefix, column.getSqlKeyword(), suffix));
+    Preconditions.checkNotNull(column);
   }
 }
--- a/src/java/com/rapleaf/jack/queries/where_operators/Match.java
+++ b/src/java/com/rapleaf/jack/queries/where_operators/Match.java
@@ -10,6 +10,6 @@ public class Match extends WhereOperator<String> {
   }
 
   public Match(Column<String> column, String prefix, String suffix) {
-    super(String.format("LIKE CONCAT('%s', %s, '%s')", prefix, Preconditions.checkNotNull(column).getSqlKeyword(), suffix));
+    super(String.format("LIKE CONCAT(CONCAT('%s', %s), '%s')", prefix, Preconditions.checkNotNull(column).getSqlKeyword(), suffix));
   }
 }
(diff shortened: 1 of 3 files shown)


src/java/com/rapleaf/jack/queries/where_operators/Match.java
```diff
@@ -1,8 +1,16 @@
...
+
 public class Match extends WhereOperator<String> {
 
   public Match(String pattern) {
     super("LIKE ?", pattern);
   }
+
+  public Match(Column<String> column, String prefix, String suffix) {
+    super(String.format("LIKE CONCAT('%s', %s, '%s')", prefix, column.getSqlKeyword(), suffix));
+    Preconditions.checkNotNull(column);
```
tuliren: I just realized that the current way we do `null` check is incorrect in a lot of the matchers. If `column` is `null`, an NPE will have been thrown in the `super` call before reaching this statement.

`checkNotNull` returns the object itself, so it is better to call it in the `super`:
```
super(..., prefix, Preconditions.checkNotNull(column).getSqlKeyword()), suffix));
```

test/java/com/rapleaf/jack/TestGenericQuery.java
```diff
@@ -708,6 +708,34 @@ public void testSimpleJoinQuery() throws Exception {
     assertEquals(commentD.getContent(), recordForCommentD.get(Comment.CONTENT));
     assertEquals(userC.getHandle(), recordForCommentD.get(User.HANDLE));
     assertEquals(postC.getTitle(), recordForCommentD.get(Post.TITLE));
+
+    results2 = db.createQuery()
+            .from(Post.TBL)
+            .leftJoin(User.TBL).on(User.ID.equalTo(Post.USER_ID.as(Long.class)))
+            .leftJoin(Comment.TBL).on(Comment.CONTENT.contains(Post.TITLE))
```
tuliren: I think this is equivalent to the following:
```java
Comment.CONTENT.contains(Post.TITLE.toString())
// or
Comment.CONTENT.contains(Post.TITLE.getSqlKeyword())
```

src/java/com/rapleaf/jack/queries/where_operators/Match.java
dwarrier: Good to know, I will add this change in.

dwarrier:
If I did:
 
`Comment.CONTENT.contains(Post.TITLE.getSqlKeyword())`

Then the generated SQL query would look like:

`... LIKE %post.title% ...`

I would get a similar result no matter what I passed in to the current `contains()` method.

because the 'contains()' method treats the argument like a string literal, not a column name. The result I want is:

`... LIKE CONCAT('%', post.title, '%') ...`

tuliren:
I see. Make sense. I misunderstood the purpose of this pull request previously.

sidoh:
Only relevant to this PR because the Travis build is failing, but we can probably update the image to use a newer version of ruby:

```
The requested url does not exist(22): 'http://rubies.travis-ci.org/ubuntu/12.04/x86_64/ree-1.8.7-2011.12.tar.bz2?rvm=1.26.10'
```

2.3.1 should be fine.

dwarrier:
FYI, I had to make a change to support both Redshift and MySQL query generation. Is this PR able to be merged?

tuliren:
Looks good. I will merge.

PR #173, opened 2016-10-05
Travis CI with Ruby 2.3

tuliren:
- Replaced java environment with ruby so that we can specify ruby version by the built-in rvm.
- Added test environment for jdk8. Each build now includes two jobs, one for jdk7, and one for 8.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -296,7 +296,7 @@
                   <goal>exec</goal>
                 </goals>
                 <configuration>
-                  <executable>/bin/sh</executable>
+                  <executable>/bin/bash</executable>
                   <arguments>
                     <argument>test/select-dbms.sh</argument>
                     <argument>mysql</argument>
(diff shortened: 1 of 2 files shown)


sidoh:
nice!

pom.xml
```diff
@@ -325,7 +325,7 @@
                   <goal>exec</goal>
                 </goals>
                 <configuration>
-                  <executable>/bin/sh</executable>
+                  <executable>/bin/bash</executable>
```
tuliren: This is necessary because `popd` and `pushd` in `select-dbms.sh` are bash builtins.

michel-tricot: Can we use sub-shells instead of these commands?

```
pushd abc
my command in abc
popd
```

```
(
cd abc
my command in abc
)

tuliren: I did not look into this issue in detail, but I guess so.

PR #174, opened 2016-11-04
Delete all lazy

bpodgursky:
@tuliren 

Since we're seeing major slowdowns in truncate speed when upgrading to 5.6/5.7, I want to see if it's faster to check if the table is empty before truncating.  Since most tests only populate 2-3 couple tables out of 550+, it will avoid the vast majority of truncates, albeit at the cost of extra queries.  

If this is significantly better, I'd probably want to replace deleteAll entirely with this implementation, but I want to test out performance first.


tuliren:
Sounds reasonable. :shipit:

PR #176, opened 2016-11-07
Add close method on db interface

tuliren:
@seancarr, I am working on adding a db connection pool, but I am not sure how soon it will be done. Meanwhile, can we use the `close` method to prevent a workflow from holding too many idle connections?


seancarr:
Looks good to me. :shipit:

seancarr:
`close` won't prevent someone from holding a connection too long but it will allow developers to close a connection when they're done with it instead of waiting for the garbage collector to sweep/close it.

tuliren:
Right. My description was imprecise.

PR #178, opened 2016-11-17
Take advantage of Java 8 features

TV4Fun:
This includes changes to take advantage of Java 8 features, as well as a lot of code cleanup and formatting changes. The external API should be unchanged but a lot of internal implementation has been updated. Major functional changes include replacing many for loops with the stream API, updating LazyLoadPersistence to use function references, adding functional objects where possible, and removing a lot of double curly brace initialization.

@bpodgursky @sidoh @tuliren @rfaugeroux @seancarr @michel-tricot


tuliren:
If there is no objection, I'd like to create a Java 7 release with the current master.

bpodgursky:
@TV4Fun is working on that right now.  We're going to do a 1.0 release of pom-common which we can do a release against.

bpodgursky:
(he's working on it for cascading_ext, but we can do the same here)

TV4Fun:
Java 7 release is done now.

TV4Fun:
How does this look? With Java 8 done and the Java 7 releases made, we should be able to merge this.

sidoh:
I think it looks fine, but would be good to run some production tests before merging.

src/java/com/rapleaf/jack/queries/IndexHints.java
```diff
@@ -5,52 +5,52 @@
...
 
-  public static IndexHint forceForOrderBy(Index index, Index... indices) {
-    return new IndexHint(IndexHint.Type.FORCE, IndexHint.Scope.ORDER_BY, index, indices);
+  public static IndexHint forceForOrderBy(Index... indices) {
+    return new IndexHint(IndexHint.Type.FORCE, IndexHint.Scope.ORDER_BY, indices);
   }
 
-  public static IndexHint forceForGroupBy(Index index, Index... indices) {
-    return new IndexHint(IndexHint.Type.FORCE, IndexHint.Scope.GROUP_BY, index, indices);
+  public static IndexHint forceForGroupBy(Index... indices) {
```
sidoh: Idea with the `T t, T... ts` pattern is to force that the user supplies at least one argument. Is that done with some other mechanism here?

src/java/com/rapleaf/jack/util/JackUtility.java
```diff
@@ -5,37 +5,23 @@
...
-    put(java.sql.Timestamp.class, new DateFormatter(TIMESTAMP_FORMATTER));
-  }};
-
-  public static final Function<Object, Long> LONG_CASTER = new Function<Object, Long>() {
-    @Override
-    public Long apply(final Object value) {
-      return Long.class.cast(value);
-    }
-  };
+  public static final Map<Class<?>, Function<Long, String>> FORMATTER_FUNCTION_MAP = new HashMap<>(2);
```
roshan: If we're willing to add a dependency on [LiveRamp commons](https://github.com/LiveRamp/commons), we can use `MapBuilder` to avoid the static initializer block.

rfaugeroux: or we can use Guava's `ImmutableMap`

src/java/com/rapleaf/jack/queries/IndexHints.java
TV4Fun: My main issue with this pattern is that it adds a lot of unnecessary complication to the code. It forces us to add extra overloads and extra lines to deal with the single arg and the varargs separately. Look for example at the constructor for WhereOperator, where we have three overloads all doing the same thing. There are also cases like in In/NotIn where we alternatively accept a Collection, and there is no way we can validate that at compile time. We could add runtime checks if you really want this validated to have at least one argument.

sidoh: I'm fine either way.

@bpodgursky @tuliren ?

tuliren: I prefer to keep the `T t, T... ts` pattern. True that in `In / NotIn` it does not prevent runtime exceptions due to empty collections, but in other class it does. In addition the code has already been written, so the cost of the complication has already been paid.

TV4Fun: Ok, I have restored the pattern to how it was.

src/java/com/rapleaf/jack/util/JackUtility.java
TV4Fun: Went with the ImmutableMap route to avoid adding an extra dependency.

TV4Fun:
Comments addressed and some other things.

sidoh:
Looks good - how can we test with existing systems?

TV4Fun:
@sidoh, besides running the code this generates on my laptop for several months without issue, I also was using it on the JDK8 slave when I was testing there.

sidoh:
I mean to suggest testing the updated generated code in a production environment.

tuliren:
@TV4Fun, sorry to see that this PR has been more than one year old. I think Jack has sufficient unit test coverage, and I think it is safe to merge it.

By the way, travis build passed, but somehow recently the build status cannot be reported back here. There is probably a permission issue.

TV4Fun:
@sidoh, I'm with @tuliren on this one. I have just merged the latest Jack changes in here and it passes all tests. I think, pending travis passing, that it's ready to merge.

sidoh:
Sounds good, definitely trust your judgment. 

I'd just suggest giving ppl a heads up and watching things carefully for a bit after merging.

PR #179, opened 2016-12-13
Update Gemfiles in test projects to use Rails 4

TV4Fun:
Otherwise bundler keeps trying to change Gemfile.lock to use Rails 3.

@bpodgursky


bpodgursky:
suuuuuuure.  fyi @tuliren

PR #180, opened 2017-01-05
Idempotent Close Method

tuliren:
@seancarr, the current `close` method is not idempotent. If this method is called twice on a connection, the second call will create a new connection and close it immediately.


seancarr:
Makes sense. :shipit:

PR #181, opened 2017-02-05
Jack Transactor

tuliren:
## Introduction
Currently Jack has no mechanism for DB connection management. When the user is not careful (e.g. if `IDb` is wrapped inside a `ThreadLocal` object, and used in a multi-threaded process with large number of threads, which was what happened in SOA Refresher), SQL operations from Jack can consume a large amount of connections from the DB server.

This problem can be solved by using a capped DB connection pool. However, a resource pool is not enough. We also need to ensure that the connection is always returned to the pool when an operation is completed. (Of course this is assuming that users cannot be trusted.) This pull request attempts to achieve this goal with a new `Transactor` class, which executes SQL operations with automatic DB management.

This pull request is far from mergeable yet. I'd like to collect some feedbacks about the interface first.

## SQL Operations
Two functional interfaces are created to specify SQL operations: `IQuery` and `IExecution`. The former has a return value and the latter does not:
```java
@FunctionalInterface
public interface IQuery<DB extends IDb, T> {
  T query(DB db) throws IOException;
}

@FunctionalInterface
public interface IExecution<DB extends IDb> {
  void execute(DB db) throws IOException;
}
```

## Transactor
A transactor runs SQL operations with a DB connection pool. There is one transactor for each `IDb`. It can be constructed as follows:
```java
ITransactor<IDatabase1> transactor = new DatabasesImpl().getDatabase1Transactor()
    .setMaxConnections(3)
    .setTimeout(Duration.standardSeconds(5))
    .get();
```

The `maxConnection` specifies the max number of DB connections allowed. The `timeout` specifies that when no idle connection is available, how long a transaction should wait before it fails due to timeout.

`Transactor` has four basic methods:
```java
public interface ITransactor<DB extends IDb> extends Closeable {
  <T> T execute(IQuery<DB, T> query);
  <T> T executeAsTransaction(IQuery<DB, T> query);
  void execute(IExecution<DB> execution);
  void executeAsTransaction(IExecution<DB> execution);
}
```
The `-AsTransaction` methods run operations in transaction.

Using Java 8 lambda, SQL operations can be executed as follows:
```java
// single line non-transactional query
long userId = transactor.query(db -> db.users().createDefaultInstance().getId());

// multi-line transactional execution
transactor.executeAsTransaction(db -> {
  db.users().find(userId).setNumPosts(111).save();
  db.posts().createDefaultInstance().setUserId(userId);
  db.comments().createDefaultInstance().setCommenterId(userId);
});
```
When there is an exception, an unchecked exception `SqlExecutionFailureException` will be thrown out. Transactor does not throw a checked exception because there is usually not much to do about it, and the code is much cleaner this way. Users can still catch `SqlExecutionFailureException` when necessary.

Commits and rollbacks will be taken care of in the `trans-` methods.

## DB Management
A DB manager inside the transactor is responsible for creating and allocating DB connections, which is delegated to `GenericObjectPool` from `commons-pool2`. The transactor gets and returns a connection from it before and after the execution of a SQL operation, respectively.

## Todo
- [x] Create unit tests.
- [x] Optimize `TransactorImpl.Builder`.

@pwestling, the previous approach we talked about actually does not work. The SQL exceptions cannot be caught to allow an automatic rollback. I think the current approach does not have that issue and is much cleaner.


src/java/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -0,0 +1,128 @@
...
+      return query.query(connection);
+    } catch (Exception e) {
+      throw new SqlExecutionFailureException(e);
+    } finally {
+      dbManager.returnConnection(connection);
+    }
+  }
+
+  @Override
+  public <T> T transQuery(IQuery<DB, T> query) {
```
bpodgursky: I think "queryAsTransaction" is only slightly wordier but more clear.

bpodgursky:
Nice.  I think this could / should wrap pretty much all our existing query code eventually.  

One question to make sure I understand -- it seems like once AbstractDbManager creates a connection, it holds onto it forever until everything is closed, right?  It would be nice if there was some kind of keep-alive time like what ThreadPoolExecutor does; otherwise the # of db connections open will be equal to the peak usage of every service.

tuliren:
@bpodgursky, yes, the `FixedDBManager` will hold onto all opened connections. My original plan is that the user is responsible for selecting a proper `maxConnections`. But it will be nice to auto close connections that have been idle for too long. I can surely add it. The only downside is that it may sometimes lead to repetitive open and close of new connections.

roshan:
Nice!

One thing, it's one connection per thread, right? The MySQL JDBC driver (IIRC) is supposed to have a thread-safe `Connection` but actually locks the same two things in two different places in two different orders, making it susceptible to deadlocking (even on read-only).

As a minor detail, `transQuery` seems to imply read-only DB operations but there's no reason the `IQuery` has to be read-only. Any reason to distinguish the two operations `transQuery` and `transExecute` in name considering they perform the same 'thing' (run against the DB in a transaction)? There's enough info there for Java to disambiguate if they're both called the same thing. I know this is contrary to what @bpodgursky says but the intent of both commands is the same, and should I change the type I'm passing in from one to the other, I gain nothing from having to change the name of the method I'm calling.

tuliren:
@roshan, `query` and `execute` are more like `callable` and `runnable`, one has a return value and the other does not. Users can define read and write operations in either of them. I totally agree with you. After reading Ben's comment, I am thinking about merging the four methods into two: `run` and `runAsTransaction`, or using another verb. Or we call it `execute` and `executeAsTransaction`, and change the naming of `IQuery` and `IExecution` to something else.

I may not have fully understood your first point. I know that if two SQL operations using the same DB connection can sometimes lead to deadlock (we had a lot of headache from this). The plan is to give each SQL operation its own connection, no matter the operation is executed in the same thread or a different one. The transactor always calls `getConnection` to grab an idle connection before executing any operation. I think that should prevent any deadlock.

roshan:
The first point is that there is internal synchronization in the JDBC driver for MySQL when `Timestamp` is used that makes it not actually compliant with the JDBC spec (AFAIK) because it is possible for it to deadlock. It doesn't look like that's a problem because of what you're doing here, though. I was being lazy and stating things without verifying that it applied to the code ;)

As for the naming, makes sense to use just the two names.

tuliren:
Method names have been unified and changed.

I do find that using the same method name to execute both `IQuery` and `IExecution` require the user to write a few more (three exactly) characters.

Suppose we change the interfaces like this:
```java
public interface ITransactor<DB extends IDb> extends Closeable {
  <T> T execute(IQuery<DB, T> query);
  <T> T executeAsTransaction(IQuery<DB, T> query);
  void execute(IExecution<DB> execution);
  void executeAsTransaction(IExecution<DB> execution);
}
```

The following query is no longer valid, because it matched both `executeAsTransaction` methods:
```java
transactor.executeAsTransaction(db -> db.users().createDefaultInstance().setNumPosts(111).save());
```

It needs to be wrapped in a block:
```
transactor.executeAsTransaction(db -> {db.users().createDefaultInstance().setNumPosts(111).save();});
```

screenshot 2017-02-05 15 36 56

tuliren:
This pull request is ready for review. @bpodgursky @roshan (or anyone), could you take a look?

src/java/com/rapleaf/jack/transaction/DbPoolManager.java
```diff
@@ -0,0 +1,153 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.rapleaf.jack.IDb;
+import com.rapleaf.jack.exception.ConnectionCreationFailureException;
+import com.rapleaf.jack.exception.NoAvailableConnectionException;
+
+class DbPoolManager<DB extends IDb> implements IDbManager<DB> {
+  private static final Logger LOG = LoggerFactory.getLogger(DbPoolManager.class);
+
+  public static int DEFAULT_MAX_TOTAL_CONNECTIONS = 1;
```
bpodgursky: Do you think as a default it would be reasonable to not limit the number here?  (-1?)

bpodgursky:
This is awesome.  :shipit:

src/java/com/rapleaf/jack/transaction/DbPoolManager.java
tuliren: I feel that if it is unlimited by default, people won't bother to adjust this parameter (similar to what happened to daemon which used to have a very short default interval to checking new configs and jammed db). Then Sean or someone else from Infrastructure will need to notify different teams to limit their db connections. What about setting it to `3` or some other small number?

bpodgursky: Makes sense.  Yeah, I think 3 or something similar makes sense.

bruce-ricard: If you want people to adjust this parameter, why don't you just make it mandatory to pass it in? You wouldn't need a default value.

tuliren: I will change it to `3` then.

Bruce, if this is a required parameter, to enable unlimited number of connections, people have to specify a negative value, which is confusing and I'd like to avoid that. The builder has an `enableInfiniteConnections` method to do that.

roshan:
Nice!

src/java/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  @Override
+  public void executeAsTransaction(IExecution<DB> execution) {
+    execute(execution, true);
+  }
+
+  private <T> T execute(IQuery<DB, T> query, boolean asTransaction) {
+    DB connection = dbManager.getConnection();
+    connection.setAutoCommit(!asTransaction);
```
seancarr: @tuliren where does autocommit get set back to it's default value (true)? When the connection is returned to the pool it could presumably still have autocommit disabled, right?

tuliren: The connection is not set back to auto commit when it is returned. The auto commit status will be set every time it is checked out of the pool and used.

npoulallion:
@roshan @bpodgursky just talked with @tuliren and @rfaugeroux about this:
I'm not a fan of having to write things like:
```
DataSyncJob job = transactor.execute((IQuery<IRlDb, DataSyncJob>)db -> db.dataSyncJobs().find(jobId));
``` 
(which can become even uglier when the query returns a list of something) or
```
DataSyncJob job = transactor.execute(db -> {
    return db.dataSyncJobs().find(jobId);
});
```
just because `execute()` can take both an `IQuery` or an `IExecution`.

I'd really love to have some `query()` method that would only accept an `IQuery` and that would look like this:
```
DataSyncJob job = transactor.query(db -> db.dataSyncJobs().find(jobId));
```
LiRen talked about it [here](https://github.com/LiveRamp/jack/pull/181#issuecomment-277560072) but never got an answer, what do you guys think?

roshan:
Fine by me. I'm not wedded to the idea if it forces ugliness at impls.

PR #182, opened 2017-02-16
Add new config options

pwestling:


pwestling:
@tuliren if you have time, can you double check me that this is fully forward compatible and won't break any use cases?

tuliren:
@pwestling, this is for docker, right? Since no interface has changed, I don't see any issue for forward compatibility.

pwestling:
Docker could be one application yes. Right now I'm working with spark and need a simpler way to transmit configs using only bash stuff.

PR #183, opened 2017-03-07
add stream method to Records

lancehc:
@tuliren It'd be nice to be able to use streams with Records without having to wrap it in a collection.


tuliren:
👍 :shipit:

PR #187, opened 2017-05-06
set the connection to null to avoid getting stuck

npoulallion:
@tuliren the goal is to avoid making a connection unusable and ending up in a bad state when trying to reset the connection when in the middle of a transaction

FYI @seancarr @rfaugeroux


src/java/com/rapleaf/jack/BaseDatabaseConnection.java
```diff
@@ -51,10 +51,11 @@ public Connection resetConnection() {
    * performed as an attempt to recover from an exception
    */
   public Connection resetConnection(Throwable cause) {
-    LOG.warn("Resetting database connection to attempt to recover from exception: {}", cause);
```
npoulallion: That's unrelated but was confusing as the brackets will always be printed in the logs.

PR #189, opened 2017-06-03
Submodules

Ongman:
This splits jack into 5 submodules:
jack-core
jack-{mysql, postgres, redshift}
jack-test

By doing so, we won't have to depend on Redshift/Postgresql if our only adapters are using Mysql, etc.

I've also added an `adapter` field to the [DatabaseDefn](https://github.com/LiveRamp/jack/compare/submodules?expand=1#diff-f5d131b73efe41ee63b32cf3557b139bR23) to simplify code generation, but it might make more sense to not do this.


jack-mysql/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+
+  <dependencies>
+
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>jack-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
```
bpodgursky: most of these deps are already included in jack-core's pom.xml, right?  shouldn't need to redefine them.

jack-mysql/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+  </properties>
+
+  <scm>
+    <connection>scm:git:git@github.com:LiveRamp/jack.git</connection>
+    <url>scm:git:git@github.com:LiveRamp/jack.git</url>
+    <developerConnection>scm:git:git@github.com:LiveRamp/jack.git</developerConnection>
+    <tag>HEAD</tag>
+  </scm>
+
+  <repositories>
```
bpodgursky: a lot of this repository and plugin configuration can go in the parent pom.xml so we don't have to duplicate it in the children

bpodgursky:
This is great.  Left some comments about pom simplification, but general structure looks good.  I think we should do a minor version bump as this is merged, to make transitioning easier.

sidoh:
Awesome!

michel-tricot:
Nice!

PR #190, opened 2017-06-06
include Gemfiles in jack-core

Ongman:
A few files from #189 were left in jack, instead of jack-core.

@bpodgursky


bpodgursky:
Does the Gemfile need to get re-added somewhere?

Ongman:
Turns out a rogue .gitignore rule excludes Gemfile in jack-core; will fix shortly.

PR #194, opened 2017-06-29
Remove deprecated transactor methods

tuliren:
@npoulallion, I think it's time to remove these methods. I will merge it after cleaning up existing use cases in the production.


npoulallion:
:fire:

PR #196, opened 2017-07-21
Stopwatch fix

amaurySabran:
The Stopwatch is causing issues with old java versions

### DIFF
--- a/jack-core/src/com/rapleaf/jack/transaction/DbMetricsImpl.java
+++ b/jack-core/src/com/rapleaf/jack/transaction/DbMetricsImpl.java
@@ -1,15 +1,13 @@
 package com.rapleaf.jack.transaction;
 
-import com.google.common.base.Stopwatch;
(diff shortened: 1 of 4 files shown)


jack-core/src/com/rapleaf/jack/transaction/DbMetricsImpl.java
```diff
@@ -31,19 +29,19 @@
   private static final Logger LOG = LoggerFactory.getLogger(DbMetricsImpl.class);
 
   public DbMetricsImpl(int maxTotalConnections, int minIdleConnections, long maxWaitMillis, long keepAliveMillis) {
-    this.lifeTimeStopwatch = new Stopwatch();
-    this.lifeTimeStopwatch.start();
+    startTime = System.currentTimeMillis();
```
tuliren: Since all the other parameters here are referred to with `this.`, let's keep the same style for `startTime`.

PR #197, opened 2017-07-26
Bug in transactor Metrics

amaurySabran:
There was a null pointer exception in the transactor metrics priority queue.
I assume this was due to a synchronization issue, I handled it and handled the null exception separately.

### DIFF
--- a/jack-core/src/com/rapleaf/jack/transaction/DbMetricsImpl.java
+++ b/jack-core/src/com/rapleaf/jack/transaction/DbMetricsImpl.java
@@ -118,7 +118,7 @@ public class DbMetricsImpl implements DbMetrics {
     summary += String.format("\nAverage number of Active connections : %,.2f", getAverageActiveConnections());
     summary += ("\nTotal number of queries/executions : " + getTotalQueries());
     summary += ("\nConnections created : " + getCreatedConnectionsCount());
-    summary += String.format("\n Max capacity time (%%) : %,.2f %%", 100 * getMaxConnectionsProportion());
+    summary += String.format("\n Max capacity time (%%) : %,.2f %%", 100 * getMaxConnectionsProportion());//TODO refactor the names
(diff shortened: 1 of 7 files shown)


jack-core/src/com/rapleaf/jack/transaction/DbMetricsImpl.java
```diff
@@ -118,7 +118,7 @@ public String getSummary() {
     summary += String.format("\nAverage number of Active connections : %,.2f", getAverageActiveConnections());
     summary += ("\nTotal number of queries/executions : " + getTotalQueries());
     summary += ("\nConnections created : " + getCreatedConnectionsCount());
-    summary += String.format("\n Max capacity time (%%) : %,.2f %%", 100 * getMaxConnectionsProportion());
+    summary += String.format("\n Max capacity time (%%) : %,.2f %%", 100 * getMaxCapacityProportion());
```
tuliren: I noticed that there are still some format inconsistency. For example, there is an extra space before the "Max". That's also the case for `TransactorMetricsImpl` line 81 (before "Average"). Can you fix them?

jack-core/src/com/rapleaf/jack/transaction/TransactorMetricsImpl.java
```diff
@@ -50,6 +54,8 @@ void update(long executionTime, StackTraceElement queryStackTrace) {
       if (longestQueriesMap.size() < longestQueriesSize) {
         longestQueriesMap.put(queryStackTrace, newQuery);
         longestQueries.add(newQuery);
+      } else if (longestQueries.isEmpty()) {
+        LOG.error("synchronization issue : longestQueries shouldn't be empty if longestQueriesMap.size()>=longestQueriesSize");
```
tuliren: Since the `update` is already synchronized, this should never happen, right?

amaurySabran: right

PR #198, opened 2017-10-06
Add SELECT DISTINCT support

JacobCrofts:
Adds a new generic query method, `selectDistinct`, which results in a `SELECT DISTINCT` query instead of `SELECT`. This is necessary for filtering out duplicate rows in queries with joins.

### DIFF
--- a/jack-test/test/java/com/rapleaf/jack/queries/TestGenericQuery.java
+++ b/jack-test/test/java/com/rapleaf/jack/queries/TestGenericQuery.java
@@ -926,4 +926,21 @@ public class TestGenericQuery {
       // expected
     }
   }
+
+  @Test
+  public void testSelectDistinct() throws Exception {
+    String selectStatement = db.createQuery()
+        .from(User.TBL)
+        .select(User.TBL.getAllColumns())
+        .getQueryStatement();
+
+    String selectDistinctStatement = db.createQuery()
+        .from(User.TBL)
+        .selectDistinct(User.TBL.getAllColumns())
+        .getQueryStatement();
+
+    assertFalse(selectStatement.contains("SELECT DISTINCT"));
+    assertTrue(selectDistinctStatement.contains("SELECT DISTINCT"));
+  }
+
 }
--- a/jack-test/test/java/com/rapleaf/jack/queries/TestGenericQuery.java
+++ b/jack-test/test/java/com/rapleaf/jack/queries/TestGenericQuery.java
@@ -936,7 +936,7 @@ public class TestGenericQuery {
 
     String selectDistinctStatement = db.createQuery()
         .from(User.TBL)
-        .selectDistinct(User.TBL.getAllColumns())
+        .select(User.TBL.getAllColumns()).distinct()
         .getQueryStatement();
 
     assertFalse(selectStatement.contains("SELECT DISTINCT"));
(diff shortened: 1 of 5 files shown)


JacobCrofts:
@LaneRamp

tuliren:
@JacobCrofts, nice! Can you add some unit tests?

JacobCrofts:
@tuliren sure thing! I'm having some IDE problems which prevent the tests from running but I'll get those sorted out ASAP

tuliren:
You are probably trying to run the jack tests under the liveramp project, which is using a working directory (for unit test) other than jack's directory (cascading_ext). That directory does not have the correct db credential file. You can do either of the three:
- Run test using maven commandline.
- Set liveramp project's working directory to jack. This is not recommended because when you run other tests, you need to revert it.
- Open jack in a new intellij project that has only one module, and then set the working directory to jack.

JacobCrofts:
When running via the command line, it complains about rake being missing. Adding rake to the gemfile results in another issue, but adding the gem should be unnecessary in the first place since nobody else has had to do this.

The specific error I get upon running tests is this:
`Error:java: Annotation processing is not supported for module cycles. Please ensure that all modules from cycle [test,jack-store] are excluded from annotation processing`

I found an outdated Stack Overflow [thread](https://stackoverflow.com/questions/27223917/how-to-configure-annotations-processing-in-intellij-idea-14-for-current-project) with a vague description of the issue, but no explicit instructions on how to resolve the dependency problem.

Would it be helpful for us to meet in person briefly tomorrow?

tuliren:
I have seen the rake error before (due to improperly set environment variable), but not the maven one.  What command are you running? Is it `mvn test -Pmysql`? You can also try to do this before the maven command:
```sh
export RAILS_ENV=development
unset BUNDLE_GEMFILE
```
Basically this is following `.travis.yml`.

JacobCrofts:
Thank you! `mvn test -Pmysql` worked. I was not specifying mysql.

jack-test/test/java/com/rapleaf/jack/queries/TestGenericQuery.java
```diff
@@ -926,4 +926,21 @@ public void testIndexHints() throws Exception {
...
+  @Test
+  public void testSelectDistinct() throws Exception {
+    String selectStatement = db.createQuery()
+        .from(User.TBL)
+        .select(User.TBL.getAllColumns())
+        .getQueryStatement();
+
+    String selectDistinctStatement = db.createQuery()
+        .from(User.TBL)
+        .selectDistinct(User.TBL.getAllColumns())
```
tuliren: The current syntax support multiple `select` method calls. With the new method, it is possible to write the following code:
```java
.select(col1, col2)
.selectDistinct(col3)
```

This is as if `col1` and `col2` are non-distinct, while `col3` is distinct, which is kind of confusing.

What about a plain `selectDistinct` method that takes no argument? When this method is called, it just flips the `this.selectDistinct` variable:
```java
.select(col1, col2)
.select(col3)
.selectDistinct()
```

JacobCrofts:
@tuliren How does this look? I decided to call the method `distinct()` because that's how the ActiveRecord equivalent is named.

tuliren:
Awesome!

PR #199, opened 2017-10-13
add getters for username and password

bpodgursky:
@tuliren on a scale of 1-10, how bad is this?  I'd like to hijack the DatabaseConnection class to get credentials from config/database.yml, but use them for some separate jetty initialization, and it would be a nuisance to try to get it to use the connection provided here.


bpodgursky:
nevermind, @pwestling abstracted this all into DatabaseConnectionConfiguration.loadFromEnvironment and I can just use that.

PR #200, opened 2017-10-24
Ignore Null When Inserting New Row

tuliren:
@bpodgursky, I guess this can prevent the exception throwing from newly added optional fields if you only create new models, instead of creating the basic model first and update the optional fields.

Unit tests have not been added yet.

### DIFF
--- a/jack-test/test/java/com/rapleaf/jack/test_project/database_1/impl/BaseCommentPersistenceImpl.java
+++ b/jack-test/test/java/com/rapleaf/jack/test_project/database_1/impl/BaseCommentPersistenceImpl.java
@@ -215,8 +215,8 @@ public class BaseCommentPersistenceImpl extends AbstractDatabaseModel<Comment> i
   }
 
   @Override
-  protected void setAttrs(Comment model, PreparedStatement stmt) throws SQLException {
-    if (model.getContent() == null) {
+  protected void setAttrs(Comment model, PreparedStatement stmt, boolean setNull) throws SQLException {
+    if (setNull && model.getContent() == null) {
       stmt.setNull(1, java.sql.Types.CHAR);
     } else {
       stmt.setString(1, model.getContent());
--- a/jack-test/test/java/com/rapleaf/jack/test_project/database_1/impl/BaseCommentPersistenceImpl.java
+++ b/jack-test/test/java/com/rapleaf/jack/test_project/database_1/impl/BaseCommentPersistenceImpl.java
@@ -23,6 +23,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.util.function.Supplier;
 
 import com.rapleaf.jack.AbstractDatabaseModel;
 import com.rapleaf.jack.BaseDatabaseConnection;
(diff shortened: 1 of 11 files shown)


bpodgursky:
Oh, that makes sense, simple enough.  Sweet, thanks : )

tuliren:
Haha, the original change was incorrect. A more correct version is actually uglier. Previously the field index is fixed. If we want to skip `null` fields for insertion, the index has to be calculated on the fly.

jack-test/test/java/com/rapleaf/jack/test_project/database_1/impl/BaseCommentPersistenceImpl.java
```diff
@@ -215,22 +215,28 @@ protected void setStatementParameters(PreparedStatement preparedStatement, Where
...
   @Override
-  protected void setAttrs(Comment model, PreparedStatement stmt) throws SQLException {
-    if (model.getContent() == null) {
-      stmt.setNull(1, java.sql.Types.CHAR);
-    } else {
-      stmt.setString(1, model.getContent());
+  protected void setAttrs(Comment model, PreparedStatement stmt, boolean setNull) throws SQLException {
+    int index = 1;
+    if (setNull && model.getContent() == null) {
+      stmt.setNull(index, java.sql.Types.CHAR);
```
bpodgursky: just to be fiddly, couldn't this be a bit terser with: 

```
+      stmt.setNull(index++, java.sql.Types.CHAR);
```

tuliren:
@bpodgursky, this pull request is ready for review. Both `save` and `create` methods should omit `null` fields now for new models.

Previously the `realCreate` method took in an `AttrSetter` and an insertion statement. The insertion statement now needs to be dynamically generated to only include non-null fields, so I introduced a new interface, `StatementSetter` that can set the parameters and return the insertion statement as well.

jack-test/test/java/com/rapleaf/jack/test_project/database_1/impl/BaseUserPersistenceImpl.java
```diff
@@ -64,75 +56,129 @@ public User create(Map<Enum, Object> fieldsMap) throws IOException {
...
         }
-          stmt.setInt(3, num_posts);
-        if (some_date == null) {
-          stmt.setNull(4, java.sql.Types.DATE);
-        } else {
-          stmt.setDate(4, new Date(some_date));
+
+        nonNullFields.add("num_posts");
+        int fieldIndex2 = index++;
+        statementSetters.add(stmt -> stmt.setInt(fieldIndex2, num_posts));
```
tuliren: Using a different variable when setting the index is necessary because variables in lambda must be final.

tuliren:
I plan to merge this PR on Monday. Although I have written some unit tests (see `TestAbstractDatabaseModel.java`), this is a scary change. Any review is welcome.

bpodgursky:
It all _seems_ reasonable, although you know far better than I do : )

I'm happy to test it out as soon as you merge.

tuliren:
This pull request does introduce a behavior change. If a column is nullable and has a default value, previously when creating a new model, that column will be set to `null`, but now it will be set to the default value. Unit tests can be broken if there are assertions on those columns for newly created models. I plan not to revert this pull request, however, because the new behavior *feels* more correct. But sorry about the inconvenience.

Thanks to @girarda, who discovered this issue.

PR #201, opened 2017-10-27
small change to let projecy.yml point jack to an inflections file

bpodgursky:
@tuliren 

This is very technically incorrect in that if it's processing multiple databases, it doesn't unload the inflections in-between.  But we don't do that, so I'm inclined to not care.


tuliren:
Makes sense.

Can you add a comment somewhere in the ruby code that the inflection file won't be unload and this may cause problem? Just in case anyone ever does that for multiple databases.

bpodgursky:
Fair enough.  Pushed disclaimer.

bpodgursky:
Thanks!

PR #202, opened 2017-11-16
allow fetching large record sets as streams

pwestling:
for cases where we need to get a large number of records and operate on them, or want to fetch records lazily, this method allows fetching a block of records at a time


roshan:
This is just as a memory optimization on the client-side, not as a performance optimization, right? A LIMIT/OFFSET does not actually have any performance benefits (excepting the fact that the query is cached each time) near the top end of the range, because for the OFFSET to work, it needs to compute at least OFFSET + LIMIT rows. 

So a `SELECT * FROM table LIMIT 100` stops at 100 rows, but a `SELECT * FROM table OFFSET 100 LIMIT 100` calculates 200 rows. So that when you're at the last "page" so-to-speak, you've calculated the whole query.

Using some sort of ORDER BY on the query (and the `id` is the natural thing, if each model is guaranteed to have one) and using the ordered-by field as a cursor is the easy way to do efficient pagination.

All of this text is assuming you're going for a performance enhancement, though. If you only want to guarantee that you have at most `k` records in memory at any point, then the whole point is moot.

pwestling:
The particular thing I'm actually targeting is PacketTooBigException for cases where we're pulling a lot of records, but performance improvements were a secondary concern. So we're getting the primary concern handled as long the sql server is not sending all of those records (I assume it's not), but I did think the performance would be better than what you're describing, so this is worth addressing. Thanks for the explanation, didn't realize how the internals worked here.

tuliren:
Even if there is not for performance enhancement, I think it is still very useful. We do run into occasional issues that the returned set is too large to retrieve in one batch.

However, one concern I have is that if the query does not specify any order, there is no guarantee on the order of the returned records. So the returned stream may be incomplete or contain duplicated entries.

https://stackoverflow.com/questions/6314879/does-limit-offset-length-require-order-by-for-pagination

jack-core/src/com/rapleaf/jack/queries/QueryFetcher.java
```diff
@@ -26,13 +32,19 @@ public static Records getQueryResults(PreparedStatement preparedStatement, Set<C
       return results;
     } catch (SQLRecoverableException e) {
       dbConnection.resetConnection();
-      throw e;
+      throw new RuntimeException(e);
```
tuliren: I think some downstream code retries if it is a sql recoverable exception. Changing it to runtime exception will disable those retires.

jack-core/src/com/rapleaf/jack/queries/GenericQuery.java
```diff
@@ -24,6 +26,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(GenericQuery.class);
   protected static int MAX_CONNECTION_RETRIES = 1;
 
+
```
tuliren: Nitpick: extra newline.

jack-core/src/com/rapleaf/jack/queries/RecordIterator.java
```diff
@@ -0,0 +1,99 @@
...
+    } catch (SQLException e) {
+      LOG.error("RecordIterator close operation failed", e);
+    }
+  }
+
+  public QueryStatistics getQueryStatistics() {
+    return queryStatistics;
+  }
+
+}
```
tuliren: Nitpick: two files missing newline at the end.

jack-core/src/com/rapleaf/jack/queries/QueryFetcher.java
pwestling: Oh, didn't mean to change this. Must have leaked from an earlier version.

PR #206, opened 2018-01-06
Sub Query

tuliren:
Summary
====

As promised to @bpodgursky and many others, Jack will support sub query by the end of this quarter. This pull request adds three new methods in `GenericQuery` such that it can be used as sub query in another `GenericQuery`:
- `GenericQuery#asSingleValue` - works for single value operators (e.g. `equalTo`, `greaterThan`) in `WHERE` clause:
  ```java
  SingleValue<Long> userId = db.createQuery()
      .from(User.TBL)
      .where(User.BIO.equalTo("B"))
      .asSingleValue(User.ID);

  records = db.createQuery()
      .from(User.TBL)
      .where(User.ID.equalTo(userId))
      .fetch();
  ```

- `GenericQuery#asMultiValue` - works for multi value operators (`IN` and `NOT IN`) in `WHERE` clause:
  ```java
  MultiValue<Long> userIds = db.createQuery()
      .from(User.TBL)
      .where(User.BIO.notEqualTo("A"))
      .asMultiValue(User.ID);

  Records records = db.createQuery()
      .from(User.TBL)
      .where(User.ID.in(userIds))
      .orderBy(User.ID)
      .fetch();
  ```

- `GenericQuery#asSubTable` - works in `FROM` clause
  ```java
  SubTable subTable = db.createQuery()
    .from(User.TBL)
    .where(User.SOME_DATETIME.notEqualTo(DateTime.parse("2017-09-10").getMillis()))
    .asSubTable("subTable");

  Records records = db.createQuery()
    .from(subTable)
    .fetch();

  // get back model in sub query
  User user = record.getModel(User.TBL.alias("subTable"), db.getDatabases()))
  // or
  User user = record.getModel(subTable.model(User.TBL), db.getDatabases()))
  ```

**See [`TestSubQuery.java`](https://github.com/LiveRamp/jack/pull/206/files#diff-0f3216fc6df1225e70a5c6dd4b61a2e2) for more examples.**

Implementation
====
The new methods convert `GenericQuery` into `SingleValue`, `MultiValue` and `SubTable` respectively. The former two extends `ValueExpression` interface, which represents a SQL query that will evaluates to some typed value. `SubTable` extends `Table` interface. It basically wrapps up a `GenericQuery` object.

Notes
====
- It is possible to eliminate `SingleValue` and `MultiValue` in the implementation. However, I think adding these two classes will make user to be more aware of the type of the sub queries they write.
- If multiple rows are returned in the `SingleValue` sub query, a runtime exception will be thrown from SQL complaining that the result should only contain one row. There is no way to add compile time check for this kind of error.
- There are totally three types of sub queries in SQL: in `select`, `from`, and `where` clause. The `select` sub query is not supported in this pull request yet. I feel that kind of sub query is not as useful as the latter two. But I can work on that if anyone feels that that is necessary.
- There are lots of corner cases in SQL sub query. I am pretty sure many of them are not covered by the current unit tests. Please help me find more of them. **Any review is welcome.**

### DIFF
--- a/jack-core/src/com/rapleaf/jack/queries/AbstractExecution.java
+++ b/jack-core/src/com/rapleaf/jack/queries/AbstractExecution.java
@@ -19,6 +19,9 @@ public abstract class AbstractExecution {
     this.dbConnection = dbConnection;
   }
 
+  /**
+   * @return fully prepared SQL statement
+   */
   public String getSqlStatement() {
     return this.getPreparedStatement(Optional.empty()).toString();
   }
@@ -26,8 +29,8 @@ public abstract class AbstractExecution {
   protected PreparedStatement getPreparedStatement(Optional<Integer> options) {
     PreparedStatement preparedStatement;
     preparedStatement = options
-        .map(integer -> dbConnection.getPreparedStatement(getQueryStatement(), integer))
-        .orElseGet(() -> dbConnection.getPreparedStatement(getQueryStatement()));
+        .map(integer -> dbConnection.getPreparedStatement(getRawStatement(), integer))
+        .orElseGet(() -> dbConnection.getPreparedStatement(getRawStatement()));
     setStatementParameters(preparedStatement, getParameters());
     return preparedStatement;
   }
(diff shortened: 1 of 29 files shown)


bpodgursky:
This is awesome!  I'm more than happy to find some places to test it out, once merged.

PR #209, opened 2018-01-30
changing interfaces to accept collections

LaneRamp:
this probably needs some cleanup, but making pr for tracking purposes and for thoughts


LaneRamp:
tests seem to be passing for me if i'm running the right ones. anything more i should do @tuliren ?

jack-core/src/com/rapleaf/jack/queries/ModelQuery.java
```diff
@@ -1,10 +1,6 @@
 package com.rapleaf.jack.queries;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
tuliren: Nit-pick: I think our IntelliJ code formatting script would actually expand `*` to specific imports. Can you run that and keep the original import style?

LaneRamp:
@tuliren lemme know if it looks good to go

tuliren:
@LaneRamp LGTM

LaneRamp:
@roshan sorry for letting the ball drop on this for so long. yeah, that is only do to reordering from formatting script (which i ran for the whole project to save time)

PR #214, opened 2018-04-11
change the order of preference for fetching config files

pwestling:
Previously, the order for fetching config files was reveresed from "normal". With this change, Jack will
first look for an entire config file stored as a string in the appropriate env var or system property, then for a path to a
config file stored in the appropriate env var or system property, then in default locations.


sidoh:
👍 looks good to me.

PR #216, opened 2018-05-03
Evaluate schema.rb instead of parsing it like a string

mharris717:
Only todo is honor FORBIDDEN_FIELD_NAMES, although no such columns appear in our schema.rb. I diffed the output from this against the old output and it was identical.

### DIFF
--- a/jack-core/src/rb/schema_rb_parser.rb
+++ b/jack-core/src/rb/schema_rb_parser.rb
@@ -12,49 +12,118 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'fattr'
+
 FORBIDDEN_FIELD_NAMES = ["tbl", "id"]
 
-class SchemaRbParser
-  extend HashRegexHelpers
+module FromHash
+  def from_hash(ops)
+    ops.each do |k,v|
+      send("#{k}=",v)
+    end
+    self
+  end
+  def initialize(ops={})
+    from_hash(ops)
+  end
+end
 
-  def self.parse(schema_rb, ignored_tables = [])
-    file_lines = File.read(schema_rb).split("\n")
-    file_lines.reject{|l| l =~ /^\s*$/}
+module ActiveRecord
+  class Schema
+    def self.define(ops = {}, &b)
+      $schema = s = new(ops)
+      s.instance_eval(&b)
+      s
+    end
 
-    models = []
-    migration_number = nil
+    include FromHash
+    fattr(:indexes) { [] }
+    fattr(:tables) { [] }
+    attr_accessor :version
 
-    until file_lines.empty?
-      line = file_lines.shift
-      if line =~ /ActiveRecord::Schema.define/
-        migration_number = extract_numeric_hash_value(line, :version)
-      elsif line =~ /create_table/ && line !~ /schema_info/
-        model_defn = ModelDefn.new(line, migration_number)
-        next if ignored_tables.include?(model_defn.table_name)
+    def create_table(name, ops = {}, &b)
+      table = Table.new(ops.merge(name: name, schema: self))
+      table.instance_eval(&b)
+      self.tables << table
+    end
 
-        ordinal = 0
-        line = file_lines.shift
-        while line =~ /^\s*t\.[a-z]+ / && !file_lines.empty?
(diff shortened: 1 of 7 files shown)


joshk0:
looks like a bunch of merge conflicts?

> We should produce schema.rb from rails 4 or rails 5 and confirm this continues to work

Sounds like a unit test to me :muscle:

joshk0:
who should be reviewing this?

roshan:
@tuliren

mharris717:
I'll clean up the stuff we called out today. A for real today, not a fake today that takes 2 weeks like last time :(

joshk0:
how's it going?

tuliren:
I think there are still unresolved TODOs, like adding support for `FORBIDDEN_FIELD_NAMES` (and maybe run travisci on different gemfile versions as mentioned in above comments). Will take a look after those are done.

joshk0:
@tuliren 
Since you are more familiar with the scenarios that Jack needs to be tested on, and the testing ecosystem, could you perhaps pair with @mharris717 on the remaining efforts? I am eager for this change to go in.

tuliren:
Currently we don't have many test cases for the ruby part. All specs live in here:
https://github.com/LiveRamp/jack/tree/master/jack-test/test/rb

If I am going to write the spec, I would probably add test cases to make sure all the old features in `SchemaRbParser` are still functioning in the new parser.

Not sure there is anything else we can pair on. The code still looks incomplete, it's not ready for testing yet. But we can sync more about this on slack.

tuliren:
> or why there are two gemfiles tbh

I have confirmed that the `Gemfile` and `Gemfile.lock` in [`jack-core/src`](https://github.com/LiveRamp/jack/tree/master/jack-core/src) are not actually used and can be deleted. Here is the [branch](https://github.com/tuliren/jack/tree/delete-core-gemfile). Feel free to delete them in this PR, unless you want me to merge mine (which may create more merge conflict).

mharris717:
@tuliren PTAL

tuliren:
@mharris717, thanks for the update. Can you add some rspecs in [jack-test/test/rb](https://github.com/LiveRamp/jack/tree/master/jack-test/test/rb)? This PR does not change anything in `db_schemas`, right?

mharris717:
@tuliren it doesn't change anything in db_schemas. 

Gonna be honest with you. Normally, I'd be the first one to be screaming to add tests for some code that didn't have it. Here, I just don't have the energy right now to add tests to a project that currently has literally one in order to service my change. I diff'd the java code generated off master against what's generated by this, and it's identical. Is that enough for you right now?

joshk0:
That's a perfectly fine approach. You've also explained the easiest test
you could write that would be really effective. Just a regression test that
compares the generated Java vs the checked in static version. That way
you're aware whenever the Java output changes, and more importantly, it's
obvious when it doesn't change.

If you don't feel like giving it a shot I think I will.

Thanks for all this! Let me know when I can sweep ruby 1.8 😀

jack-core/src/rb/schema_rb_parser.rb
```diff
@@ -12,49 +12,118 @@
...
+        res[f.to_s] = v if !v.nil?
       end
+      res
     end
-    return models, migration_number
+  end
+end
+
+class SchemaRbParser
+  def self.parse(schema_rb, ignored_tables)
```
tuliren: `ignored_tables` is no longer optional, which breaks the main function (line 131, post your change).

tuliren: This is fixed in https://github.com/LiveRamp/jack/commit/a2617c8e26d8e0f11ae10fe92cc9dde7dff8f1d3.

joshk0:
Thanks for this @tuliren . How would we know when this version of jack is in use throughout liveramp and we can sweep the old version of Ruby?

tuliren:
Currently the redshift schema project is broken due to this change. I am working on a fix. Other than that, since jenkins build already succeeded for jack, I suppose it has been propagated to all projects company wide.

tuliren:
`redshift_schemas` has been fixed. I am pretty sure the new version of jack is used everywhere now. But I don't know where the old ruby versions are used.

PR #217, opened 2018-05-04
add flags for ResultSet type and concurrency to enable row-by-row streaming of results correctly

pwestling:


joshlang232:
@pwestling Looks good.

From https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-implementation-notes.html:
```
The combination of a forward-only, read-only result set, with a fetch size of Integer.MIN_VALUE serves as a signal to the driver to stream result sets row-by-row. After this, any result sets created with the statement will be retrieved row-by-row.
```

PR #219, opened 2018-06-01
use java.time.Duration in setMaxWaitTime and setKeepAliveTime and dep…

npoulallion:
…recate old methods + update tests

@tuliren any reason why we use joda time instead of the native java.time?
If not, what do you think about that?


npoulallion:
Oh, that makes sense!
Thanks for the quick review :)

PR #236, opened 2019-02-19
Support aggregated column in subquery

tuliren:
@mark-idleman, this PR should allow us to use aggregated columns in subqueries and fix your query issue. An example is in `testAggregatedColumnSubQuery`.

The main idea is that the sql expression of a column should be different for different clauses:
1. Normally it is just the column name.
2. In select clause, when an alias is need, it is in the form of "column AS alias". This is necessary for aggregated columns within a subquery.
3. When retrieving the value from a `ResultSet`, it is just the alias.

I've compared a few implementations, and this one is the shortest that requires the least amount of refactoring.

### DIFF
--- a/jack-core/src/com/rapleaf/jack/queries/AggregatedColumn.java
+++ b/jack-core/src/com/rapleaf/jack/queries/AggregatedColumn.java
@@ -7,10 +7,21 @@ public class AggregatedColumn<T> extends Column<T> {
   }
 
   private final Function function;
+  private final String sqlKeyword;
+  private final String alias;
 
   private AggregatedColumn(Column column, Function function) {
-    super(column);
+    super(column.table, column.field, column.type);
     this.function = function;
+    this.sqlKeyword = function.toString() + "(" + column.getSqlKeyword() + ")";
+    this.alias = column.getSqlKeyword().replaceAll("\\.", "_") + "_" + function.name().toLowerCase();
+  }
+
+  private AggregatedColumn(Column column, Function function, String sqlKeyword, String alias) {
+    super(column.table, column.field, column.type);
+    this.function = function;
+    this.sqlKeyword = sqlKeyword;
+    this.alias = alias;
   }
 
   public static <T> AggregatedColumn<Integer> COUNT(Column<T> column) {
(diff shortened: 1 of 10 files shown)


mark-idleman:
Read through this and I think it all looks good to me - the tests definitely help as well, as I'm not super familiar with the Jack internals. My only question: in the `GenericQuery` class, you've added a function that you pass into the `getClauseFromColumn` function for select and group by clauses that allows for the proper column alias to be used when forming raw SQL for those SQL commands. 

Is it the case that if we wanted aggregated columns to work for other types of SQL statements (`ORDER BY`, for instance), we'd need to make similar code changes? It seems `ORDER BY` and other SQL statements use `getClauseFromQueryConditions` rather than `getClauseFromColumn`, but I'm guessing that this difference is likely due to underlying Jack details I'm currently not aware of.

tuliren:
Good point about other clauses. I think for those clauses, however, an expression is fine (e.g. we can order by `MAX(raw_version)`). But it is possible that I may have missed something. Is there any specific case you can think of that I should test?

jack-core/src/com/rapleaf/jack/queries/AggregatedColumn.java
```diff
@@ -7,10 +7,21 @@
   }
 
   private final Function function;
+  private final String sqlKeyword;
+  private final String alias;
 
   private AggregatedColumn(Column column, Function function) {
-    super(column);
+    super(column.table, column.field, column.type);
```
tuliren: @roshan, I deleted the copy constructor because there was only one usage of that constructor. I am kind of opinionated towards keeping as few overloaded methods as possible. In the future, if there is more use case, we can add it back. Does it sound good to you?

roshan: Sorry, I just mean that there's [this constructor](https://github.com/LiveRamp/jack/pull/236/files#diff-7fd887cbe87236ce962da94adf3a6578R20) which can be used here as:

```
this(
  column,
  function,
  function.toString() + "(" + column.getSqlKeyword() + ")",
  this.alias = column.getSqlKeyword().replaceAll("\\.", "_") + "_" + function.name().toLowerCase()
)
```

Right?

tuliren: Oh, right. Will fix.

PR #237, opened 2019-03-05
Monitoring DB Connection Pool Usage

henryzhao81:


jack-core/src/com/rapleaf/jack/transaction/DbPoolStatusImpl.java
```diff
@@ -0,0 +1,36 @@
...
+  public int getNumIdle() {
+    return this.connectionPool.getNumIdle();
+  }
+
+  /**
+   * @return The number of threads currently blocked waiting for an db connection object from the pool
+   */
+  @Override
+  public int getNumWaiters() {
+    return this.connectionPool.getNumActive();
```
tuliren:
```suggestion
    return this.connectionPool.getNumWaiters();
```

jack-core/src/com/rapleaf/jack/transaction/DbPoolStatus.java
```diff
@@ -0,0 +1,10 @@
+package com.rapleaf.jack.transaction;
+
+public interface DbPoolStatus {
+
+  int getNumActive();
+
+  int getNumIdle();
+
+  int getNumWaiters();
+}
```
tuliren: Nitpick: newline in this interface is not consistent. Either remove the extra newline before `int getNumActive()`, or add a newline after `int getNumWaiters()`.

PR #238, opened 2019-03-30
Create mariadb submodule

tuliren:
@jrhizor, I think a lot of the boilerplate code for mariadb can be extracted into a submodule in jack. The only thing that is needed from the user is the `InMemoryDb` class. What do you think?

### DIFF
--- a/jack-mariadb/pom.xml
+++ b/jack-mariadb/pom.xml
@@ -1,0 +1,61 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>jack-mariadb</artifactId>
+  <version>1.2-SNAPSHOT</version>
+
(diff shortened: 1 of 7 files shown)


jack-mariadb/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+  <version>1.2-SNAPSHOT</version>
+
+  <parent>
+    <groupId>com.liveramp</groupId>
+    <artifactId>jack</artifactId>
+    <version>1.2-SNAPSHOT</version>
+  </parent>
+
+  <properties>
+    <db.user>root</db.user>
```
jrhizor: These aren't necessary, right?

jack-mariadb/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+      <plugin>
+        <artifactId>exec-maven-plugin</artifactId>
+        <groupId>org.codehaus.mojo</groupId>
+        <version>1.3.2</version>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <failIfNoTests>false</failIfNoTests>
```
jrhizor: We should fail if there are no tests for this module.

jrhizor:
Closing, we're going to do a different version of db providers.

PR #240, opened 2019-04-02
Fix sonatype branch

tuliren:
There are two ways to fix it.
1. Enforce the build and deploy to use ruby `2.3.0`.
2. Upgrade `activesupport` to `4.2.8`.


tuliren:
Hmm, launching a virtual machine on travis-ci.com is really slow...

PR #241, opened 2019-04-17
add testcontainers support

jrhizor:
There might be a little more work necessary to get this to pass tests (especially for launching containers on our OSS CI).

See `TestMysqlContainerDb` for an example of what we're hoping to achieve (easy accessors to a single containerized MySQL instance per database per JVM). This instance should be initialized with the Jack schema dump.


jrhizor:
It turns out that the test module here does not add the schema dump as a Java resource like we do internally. We'll either need to get that working or remove the testing.

jrhizor:
Going to close since I'm not going to be able to commit to this in the near future. 

To actually productionize this, we could have `MysqlContainerSingletonProvider` take in a heap dump location and provide an internal wrapper that loads it from our internal schema location.

PR #245, opened 2019-05-20
Handle throwables while executing transactions

sherifnada:
## Context
Both methods which execute transactions (`query` and `execute`) only catch `Exception`s, not `Throwables`. As a result, if a `Throwable` is thrown in the middle of a transaction, then we only execute the `finally` block which returns the connection (and we don't explicitly rollback the transaction as we would have done inside the `catch(Exception e)` block. The unfortunate side effect of this behavior is that it commits partially-completed transactions; returning a connection calls `.setAutoCommit(true)` which commits any commands built up during the uncommitted transaction which leaves us with inconsistent state. 

## Changes introduced in this PR
1. remove the `finally` block, and instead return the connection at the end of both the `try` and `catch(Exception e)` blocks. 
2. add a `catch(Throwable t)` block in which we try to rollback any uncommitted changes, but do not attempt to return the connection. 
3. In the interest of DRY, refactor `execute` to be a wrapper on top of `query` instead of an almost exact replica.

### DIFF
--- a/jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
+++ b/jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
@@ -1,14 +1,14 @@
 package com.rapleaf.jack.transaction;
 
-import java.time.Duration;
-import java.util.concurrent.Callable;
-
 import com.google.common.base.Preconditions;
+import com.rapleaf.jack.IDb;
+import com.rapleaf.jack.exception.SqlExecutionFailureException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.rapleaf.jack.IDb;
-import com.rapleaf.jack.exception.SqlExecutionFailureException;
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.Callable;
 
 /**
  * If there is any exception while executing the query, throws {@link com.rapleaf.jack.exception.SqlExecutionFailureException}.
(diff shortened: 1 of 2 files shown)


jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -85,44 +86,40 @@ DbMetrics getDbMetrics() {
...
-      throw new SqlExecutionFailureException(e);
-    } finally {
-      // connection should be reset in PooledObjectFactory
-      dbManager.returnConnection(connection);
+
+      throw t;
     }
   }
 
+  private void execute(final IExecution<DB> execution, boolean asTransaction) {
```
sherifnada: the diff looks a little weird, but I replaced the previous `execute` method with a call to `query`. The two methods are identical except for the fact that `execute` has `void` return type and `query` returns an object. So I just made `execute` a thin wrapper on top of `query`.

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -85,44 +86,43 @@ DbMetrics getDbMetrics() {
...
+
+      throw t;
     }
   }
 
+  private void execute(IExecution<DB> execution, boolean asTransaction) {
+    query(
+        db -> {
+          execution.execute(db);
+          return Optional.empty();
```
tuliren:
```suggestion
          return null;
```
Nitpick: returning `null` also works, and is simpler.

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -21,6 +21,7 @@
  * If new DB connections cannot be created, throws {@link com.rapleaf.jack.exception.ConnectionCreationFailureException}.
  */
 public class TransactorImpl<DB extends IDb> implements ITransactor<DB> {
+
```
tuliren: Nitpick: this extra newline is unnecessary.

roshan:
Good catch, @sherifnada . Nice work!

PR #248, opened 2019-07-19
added getConnection() retry policy

ziqizh:
Current `getConnection()` function immediately throws exception if the SQL server is unreachable. As we migrated to CloudSQL, it needs a retry policy to deal with the downtime of CloudSQL. More details see: [Aim service retries when connecting to bang DB](https://liveramp.atlassian.net/browse/DEVID-997)

### DIFF
--- a/jack-core/src/com/rapleaf/jack/DatabaseConnection.java
+++ b/jack-core/src/com/rapleaf/jack/DatabaseConnection.java
@@ -103,7 +103,7 @@ public class DatabaseConnection extends BaseDatabaseConnection {
    * If the connection hasn't been used in a long time, close it and create a new one.
    * We do this because MySQL has an 8 hour idle connection timeout.
    * <p>
-   * Because of the potential downtime of CloudSQL, it implements exponential retry policy.
+   * Because of the intermittent downtime of SQL service, it implements exponential retry policy.
    * The default retry policy retries seven times, which handles SQL downtime less than approx. two minutes.
    */
 
--- a/jack-core/src/com/rapleaf/jack/DatabaseConnection.java
+++ b/jack-core/src/com/rapleaf/jack/DatabaseConnection.java
@@ -36,6 +36,7 @@ import static com.rapleaf.jack.DatabaseConnectionConstants.REDSHIFT_JDBC_DRIVER;
  */
 public class DatabaseConnection extends BaseDatabaseConnection {
   private static final String PARTITION_NUM_ENV_VARIABLE_NAME = "TLB_PARTITION_NUMBER";
+  private static final int GET_CONNECTION_MAX_RETRIES = 7;
 
   static final Map VALID_ADAPTER_CONSTANTS = Collections.unmodifiableMap(new HashMap() {
     private static final long serialVersionUID = 1L;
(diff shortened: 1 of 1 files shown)


jack-core/src/com/rapleaf/jack/DatabaseConnection.java
```diff
@@ -101,21 +101,39 @@ public DatabaseConnection(String dbname_key, long expiration, String driverClass
...
+          conn = DriverManager.getConnection(connectionString, username.orNull(), password.orNull());
+        } else if (isExpired() || conn.isClosed()) {
+          resetConnection();
+        }
+        updateExpiration();
+        return conn;
+      } catch (Exception e) { //IOEx., SQLEx.
+        // if it is the last retry, throw exception
+        if (retryCount == MAX_RETRIES - 1) {
+          throw new RuntimeException(e);
```
bvivio: I think including some useful error message in this `RuntimeException` will make debugging and/or log parsing a little easier. I'm thinking something along the lines of
```
throw new RuntimeException(
    String.format("Could not establish connection after %d attempts", maxNumRetries),
    e
);
```

jack-core/src/com/rapleaf/jack/DatabaseConnection.java
```diff
@@ -101,21 +101,39 @@ public DatabaseConnection(String dbname_key, long expiration, String driverClass
...
+          // Catches interruptions in the sleep
+        }
       }
-      updateExpiration();
-      return conn;
-    } catch (Exception e) { //IOEx., SQLEx.
-      throw new RuntimeException(e);
     }
+    // it should never reach at this point
+    return null;
```
bvivio: If the code should never reach this point, then perhaps just throwing a `RuntimeException` (with a useful error message) would be better than returning `null`? Seems like failing hard and fast would make debugging easier than discovering a mysterious `NullPointerException` somewhere else in the stacktrace as a result of this method returning `null`.

jack-core/src/com/rapleaf/jack/DatabaseConnection.java
```diff
@@ -101,21 +101,39 @@ public DatabaseConnection(String dbname_key, long expiration, String driverClass
    * Get a Connection to a database. If there is no connection, create a new one.
    * If the connection hasn't been used in a long time, close it and create a new one.
    * We do this because MySQL has an 8 hour idle connection timeout.
+   *
+   * Because of the potential downtime of CloudSQL, it implements exponential retry policy.
```
bvivio: nit: perhaps a more general description about handling intermittent MySQL downtime would be more broadly applicable than a statement regarding the specific database service that LiveRamp uses (CloudSQL)?

jack-core/src/com/rapleaf/jack/DatabaseConnection.java
```diff
@@ -101,21 +101,39 @@ public DatabaseConnection(String dbname_key, long expiration, String driverClass
...
    */
 
   public Connection getConnectionInternal() {
-    try {
-      if (conn == null) {
-        Class.forName(driverClass);
-        conn = DriverManager.getConnection(connectionString, username.orNull(), password.orNull());
-      } else if (isExpired() || conn.isClosed()) {
-        resetConnection();
+    int MAX_RETRIES = 7;
```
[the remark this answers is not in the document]
ziqizh: Thank you for your comments! I changed MAX_RETRIES to a global constant and set it to 7 by default. I also think I can add a setter for MAX_RETRIES if necessary.

jack-core/src/com/rapleaf/jack/DatabaseConnection.java
```diff
@@ -101,21 +102,38 @@ public DatabaseConnection(String dbname_key, long expiration, String driverClass
...
+        // if it is the last retry, throw exception
+        if (retryCount == GET_CONNECTION_MAX_RETRIES - 1) {
+          throw new RuntimeException(String.format("Could not establish connection after %d attempts",
+              GET_CONNECTION_MAX_RETRIES), e);
+        }
+        // Wait (2^retryCount) seconds
+        try {
+          Thread.sleep(Math.round(Math.pow(2, retryCount)) * 1000);
+        } catch (Exception ignored) {
+          // Catches interruptions in the sleep
```
jrhizor: Can you store this exception and include it as part of the final `RuntimeException` you throw? Otherwise, we will know it failed to create a connection will be known but we won't know why.

ziqizh: The program is not supposed to reach final `RuntimeException` because it should either successfully create a connection or fail seven times and throw a `RuntimeException` in the 7th try, which includes the detail of the `exception`. 
In this case, do we still need to store this exception?

p.s. The try-catch block of `Thread.sleep` is to ensure the retry continues if the process got interrupted during sleep, and I didn't use the exception since it doesn't contain any interesting information;

jrhizor: Oh, sorry. I completely misread.

ziqizh: nvm I think it is better to include the information in the final `runtimeException` in case it reaches to the final one.

PR #258, opened 2019-09-17
Fix connection leak caused by exceptions during rollback.

briancecker:
If an exception is thrown while attempting to rollback a transaction,
then the connection is never returned to the connection pool. This
slowly leaks connections until there are none left.

### DIFF
--- a/jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
+++ b/jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
@@ -96,7 +96,15 @@ public class TransactorImpl<DB extends IDb> implements ITransactor<DB> {
     } catch (Exception e) {
       LOG.error("SQL execution failure", e);
       if (asTransaction) {
-        connection.rollback();
+        try {
+          connection.rollback();
+        } catch (Exception e2) {
+          LOG.warn("Failed to rollback an active transaction", e2);
+          // ignore. We can't actually do anything about this, and we need to return the connection
+          // to the pool under all circumstances where the application will continue running
+          // otherwise we will slowly leak connections until all connections are exhausted from the
+          // pool and will never be returned.
+        }
       }
 
       dbManager.returnConnection(connection);
--- a/jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
+++ b/jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
@@ -7,7 +7,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
-import java.util.Optional;
 import java.util.concurrent.Callable;
 
 /**
@@ -74,6 +73,8 @@ public class TransactorImpl<DB extends IDb> implements ITransactor<DB> {
 
   private <T> T query(IQuery<DB, T> query, boolean asTransaction) {
     DB connection = dbManager.getConnection();
+    boolean connectionSafeToReturn = true;
+
     try {
       connection.setAutoCommit(!asTransaction);
       long startTime = System.currentTimeMillis();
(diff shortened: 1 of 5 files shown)


briancecker:
The heap dump below shows the state of a connection pool that has been completely allocated out. 
pl_id_syncer_connections

If we further inspect these allocated objects, we can see that in our case, they were borrowed nearly two days ago, and the logs show that at exactly the time when they were last borrowed they threw an `Exception` during a transaction, and then failed to ever execute a DB query again. 
Screen Shot 2019-09-17 at 10 16 23 AM

The epoch time for `lastBorrowTime` corresponds to September 15, 2019 11:23:37, exactly the time in our logs where we start to see consistent errors until restart. 
Screen Shot 2019-09-17 at 10 18 33 AM

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -96,7 +96,14 @@ DbMetrics getDbMetrics() {
     } catch (Exception e) {
       LOG.error("SQL execution failure", e);
       if (asTransaction) {
-        connection.rollback();
+        try {
+          connection.rollback();
```
joshcurtis: Does it make sense to put this in a `try-finally` block? And remove this block below since the connection was returned?
``` java
if (asTransaction) {
  connection.rollback();
}
```

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
```diff
@@ -96,7 +96,14 @@ DbMetrics getDbMetrics() {
...
+        } catch (Exception e2) {
+          // ignore. We can't actually do anything about this, and we need to return the connection
```
h-wang94: Is it worth logging that we were unable to rollback?

jack-test/test/java/com/rapleaf/jack/transaction/TestDbTransactorImpl.java
```diff
@@ -38,6 +44,43 @@ public void cleanup() throws Exception {
...
+    DbPoolManager<IDatabase1> dbPoolManager = Mockito.spy(new DbPoolManager<>(() -> database1, 1,
+        1, 100, Integer.MAX_VALUE, false
+    ));
+    TransactorImpl<IDatabase1> transactor = new TransactorImpl<>(dbPoolManager, false);
+
+    try {
+      transactor.queryAsTransaction(db -> db.users().find(1));
+      fail();
+    } catch (Exception e) {
+      // This should fail
```
h-wang94: Probably need to fix up the wording here. It currently seems like its implying something should be failing inside this catch block instead of implying that this is the correct behavior.

jack-test/test/java/com/rapleaf/jack/transaction/TestDbTransactorImpl.java
```diff
@@ -38,6 +44,43 @@ public void cleanup() throws Exception {
...
+    }
+
+    Mockito.verify(dbPoolManager, Mockito.times(1)).returnConnection(Mockito.any(IDatabase1.class));
+
+    // This simply needs to not throw, demonstrating that the connection pool is not exhausted.
+    transactor.query(db -> db.users().find(1));
```
h-wang94: I hate a bajillion magic numbers. 

Can we perhaps assign `1` to some variable at the top of this test and then re-use it so its clear what the relationship of the values are?

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
briancecker: Do you mean remove the `catch (Throwable t)` block below and add a `finally` block? I think it makes sense to do that, but there's this big block comment above that explicitly mentions why we don't do that. I sort of disagree with that comment though, so I'm hoping @sherifnada can weigh in as he added that comment and logic.

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
briancecker: Oh, yeah that might be useful. I'll log it.

jack-test/test/java/com/rapleaf/jack/transaction/TestDbTransactorImpl.java
briancecker: Ah. Yeah, I can assign it to a variable and leave a comment. It basically has no relation. I just need to query for something. It doesn't matter that it happens to be the same id in both the queries.

jack-test/test/java/com/rapleaf/jack/transaction/TestDbTransactorImpl.java
briancecker: Changed it. Hopefully it's more clear now.

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
hswaffield: I think josh is saying we should always return the connection, regardless of how the rollback went, so it makes sense to just put that particular piece inside a `finally` block, once, and put the rollback inside a `try` block... that way we have `dbManager.returnConnection(connection);` written only once here.

I agree with that idea.

hswaffield:
I didn't previously see the comments that Sherif left before... I look forward to hearing more of his thoughts here.

briancecker:
Hmm, I've come up with a slightly different solution here. Instead of always returning the connections, how about we expose a version of the `ObjectPool#invalidateObject` method in the `IDbManager` interface as something like `IDbManager#invalidateConnection`? In this way we can explicitly destroy the connection and remove it from the pool without having to rely on any validation checks in the pool to later remove the invalid connection and without accidentally committing any half-finished transactions (although I think that is unlikely).

briancecker:
I went ahead and introduced `IDbManager#invalidateConnection` and modified `TransactorImpl` to make use of it as I mentioned above. I think the code is quite a bit cleaner now, and it makes a lot more sense logically too! Please review again if you have time. I'd love to get this merged today :)

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
briancecker: I think this should be resolved now with the recent changes.

briancecker:
> Just to confirm, by just invalidating the `<DB extends IDb> connection` we won't accidentally leave connections open on the DB side, will we? 

Good question! Calling the `DbPoolManager#invalidateConnection` eventually calls `DbPoolFactory#destroyObject`, which is defined as:
```
    @Override
    public void destroyObject(PooledObject<DB> connection) throws IOException {
      connection.getObject().close();
    }
```

So it should close the connection. Of course that `.close()` call could fail for some reason, but then we're out of options, and we just have to wait until the connection expires on the server if it somehow ended up orphaned, but I don't actually know how that would happen.

jack-core/src/com/rapleaf/jack/transaction/TransactorImpl.java
sherifnada: Hey guys, just catching up after coming back from a trip to China. It looks like this was merged already -- @briancecker did you feel that comment was inaccurate? or that it's the wrong way of fixing that problem?

hswaffield: We believe it was the wrong way to fix the problem because it resulted in the `DBPoolManager` never letting go of stale connections, in certain situations, which killed a number of our apps as they could not get any new connections.

briancecker: As Henry mentioned, we didn't so much feel it was inaccurate, but the way of solving the issue lead to another issue. To preserve the behavior you added, avoid committing half-finished transactions, and avoid leaking connections, I added a new method for invalidating a connection instead of returning it. This method (`DbPoolManager#invalidateConnection`) removes the connection from the pool's tracked objects and closes it without committing.

PR #265, opened 2020-05-08
INGEST-2133, Update mysql-connector-java

rupertchen:
The previous version being used is five years old and numerous bug fixes
have been applied since.  Chief among them is re-enabled support for
fractional seconds.

Jira: INGEST-2133

### DIFF
--- a/jack-mysql/pom.xml
+++ b/jack-mysql/pom.xml
@@ -35,7 +35,6 @@
     <dependency>
       <groupId>mysql</groupId>
       <artifactId>mysql-connector-java</artifactId>
-      <version>5.1.22</version>
     </dependency>
 
   </dependencies>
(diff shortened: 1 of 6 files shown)


rupertchen:
I introduced `pluginManagement` because all the modules were missing the maven-surefire-plugin version. Once I added that, it made sense to me that exec-maven-plugin should be the same across all modules as well.

Since both jack-mysql and jack-test used mysql-connector-java, I placed that into `dependencyManagement`.

Lastly, I removed the property `maven.compiler.source` because it was redundant with the configuration already in the plugin. I added `project.build.sourceEncoding` since that was the [recommended way to configure the encoding][1] and there were maven warnings because it was missing before.

[1]: https://maven.apache.org/plugins/maven-resources-plugin/examples/encoding.html

PR #283, opened 2024-02-23
[OC-49258] configure deploy profile

johnjelinek:


Jenkinsfile
```diff
@@ -5,5 +5,5 @@ mvnBuildPipeline {
     environment {
         JAVA_HOME = '/usr/lib/jvm/java-8-openjdk-amd64'
     }
-    mavenPropertiesOverride = '-Dmaven.test.failure.ignore=false -Ddb.user=$HUDSON_DB_USER -Ddb.pass=$HUDSON_DB_PASS -Pmysql,liveramp-ci'
+    mavenPropertiesOverride = '-Dmaven.test.failure.ignore=false -Ddb.user=$HUDSON_DB_USER -Ddb.pass=$HUDSON_DB_PASS -Pmysql'
```
johnjelinek, line 8: the profile is already included by default, so don't need to specify it here

